PR #179, opened 2019-10-08
Generate HTML instead of Markdown

brendanzab:
This generates HTML instead of Markdown from the documentation compiler. This should be easier to generate, and gives us more flexibility in the future when documenting more complicated things, like if expressions, etc.


crates/ddl/src/core/compile/doc/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+
+    const MINIRESET: &str = include_str!("./minireset.min.css");
+    const STYLE: &str = include_str!("./style.css");
+
+    let pkg_name = env!("CARGO_PKG_NAME");
+    let pkg_version = env!("CARGO_PKG_VERSION");
+
+    write!(
+        writer,
+        "\
```
wezm: Maybe use a raw string here to allow writing a bit more naturally (no need to escape quotes for example).

crates/ddl/src/core/compile/doc/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+}
+
+fn compile_alias(
+    context: &ModuleContext,
+    writer: &mut impl Write,
+    alias: &core::Alias,
+    report: &mut dyn FnMut(Diagnostic),
+) -> io::Result<()> {
+    writeln!(writer, "    <dt class=\"item alias\">")?;
+    writeln!(writer, "      <a href=\"#\">{}</a>", alias.name)?;
```
wezm: I'm always a little wary of string interpolation into XML/HTML. Seems this is probably ok given identifiers are unlikely to contain characters that need escaping in HTML. May be worth considering a template language that guarantees a bit of correctness around tag generation and escaping at some point.

crates/ddl/src/core/compile/doc/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+            }
+        }
+    }
+
+    write!(
+        writer,
+        "\
+  </dl>
+</body>
+"
```
wezm: `</html>` tag is missing.

crates/ddl/src/core/compile/doc/mod.rs
brendanzab: Yeah, I was definitely thinking of going down that route in the future!

crates/ddl/src/core/compile/doc/mod.rs
brendanzab: Oh, nice catch!

crates/ddl/src/core/compile/doc/mod.rs
brendanzab: Good idea. This might be better solved with a better templating language though... so I think I'm fine with just doing escaping for now! 😅

brendanzab: Decided to go with the raw string literals after all!

PR #181, opened 2019-10-08
Avoid caching the target directory during CI

brendanzab:
The travis caches are currently between 5 and 10Gb and the build spends a lot of time (about 6 mins x 2) downloading and unpacking before the build and packing and uploading afterwards.


brendanzab:
This seemed to speed things up, and reduced the cache size!

PR #183, opened 2019-10-10
Implement module-level doc comments

### LINKED ISSUE #155: Top level documentation
We should be able to write top level documentation in our DDL files. Eg.

- `test.ddl`

  ```
  //! This is a data definition.
  ```
- `test.rs`

  ```rust
  //! This is a data definition.

  struct Empty {}
  ```
- `test.md`

  ```md
  This is a data definition.

  ## Empty
  ```

brendanzab:
Closes #155.


brendanzab:
Sadly I need to figure out a workaround for a compiler bug in rustc.

foo.rs:
```
//! Testing simple aliases.

pub type Byte = ddl_rt::U8;
```

main.rs:
```
mod foo {
    include!("./foo.rs");
    use ddl_test_util::ddl_rt;
}

fn main() {
    println!("Hello, world!");
}
```

Here rustc complains that `//! test` is not a valid place for an outer doc comment, when it actually is. While I could do the following, I need a way to figure out a workaround for the imports:

```
#[path = "./foo.rs"]
mod foo;
```

Issue #161, opened 2019-07-25
Figure out how to carve up our terms into 'sub-universes' with certain properties


brendanzab:
Without becoming to wedded to a specific way of doing things, we might also want to do some analysis on the 'kind' of things: `Type`, `Format`, `Predicate`, `kind -> kind`. This would allow us to express the following types:

```
// Format arrays
Array : Nat -> Format -> Format
// Host lists
List : Type -> Type
```

We want to be able to do certain forms of semantic analysis on our terms. We care about properties like:

- sizedness: eg. fixed number of bytes, unknown size
- alignment: eg. byte aligned, unknown alignment

This should compose nicely with the other ways of doing things in our type system.

Now that I write this, it seems like these properties are both specific to `Format`s. Perhaps we could track these in there? Ie. having something like `Format size align`, with some defaults. Just an idea!

PR #186, opened 2019-10-16
Move to a Pure Type System and add a `Format` kind

brendanzab:
Right after I removed kinds from our type system, I’m adding them back in! This moves us to a Pure Type System (PTS), which are pretty well known in the literature:

- https://en.wikipedia.org/wiki/Pure_type_system
- https://ncatlab.org/nlab/show/pure+type+system

I knew about these before, but now I’m starting to understand them a bit better!

We use the following definitions:

- sorts = { `Type`, `Format`, `Kind` }
- axioms = { (`Type`, `Kind`), (`Format`, `Kind`) }

We don’t have any formation rules yet, due to a lack of function types.

This may help us towards: #161


brendanzab:
One of the big motivations here is that I want to restrict term-type dependency in `Types`, but allow it in between `Type`s and `Format`s. I _think_ I might be able to do this via the PTS formation rules, but I'll have to give it a think.

brendanzab:
Here is what I think we might end up with:

```rust
sorts = { Type, Format, Kind }
axioms = { (Type, Kind), (Format, Kind) }
formation rules = {
sorts = { Type, Format, Kind }
axioms = { (Type, Kind), (Format, Kind) }
formation rules = {
    (Type, Type, Type),         // eg. Int -> Int       : Type
    (Type, Format, Format),     // eg. Int -> Format    : Format
    (Type, Format, Format),     // eg. Type -> Format   : Format
    (Type, Format, Format),     // eg. Type -> Format   : Format
    (Kind, Kind, Kind),         // eg. Type -> Type     : Kind
                                //     Format -> Format : Kind
                                //     Type -> Format   : Kind
                                //     Format -> Type   : Kind
}
}
```

brendanzab:
Omg, I think this actually might work, unless I'm missing something major. 😳

brendanzab:
I think things twigged for me a number of months back when I was looking at ["Inductive Types Deconstructed: The Calculus of United Constructions"](http://www.iro.umontreal.ca/~monnier/itd-tyde-2019.pdf), where show the calculus of constructions with universes as a PTS:

Screen Shot 2019-10-16 at 4 29 28 pm

Was super interesting to see this creative use of the sets of sorts, axioms, and formation rules. Then I just needed t go back over the exisiting ground I had visited with regards to the DDL.

brendanzab:
I think I'll merge this. We can always back out later on if we decide this is not the right way to go! At any rate, I'm now happier with how kinds are integrated in comparison to what we used to have before I removed them.

Issue #190, opened 2019-10-26
Simple DDL compilation strategy for Rust


mikeday:
To compile the DDL to Rust we need a compilation strategy that is _complete_ in that it can handle every binary format type, _simple_ in that it is easy to implement, and _usable_ in that it presents a reasonable interface to Rust code that wishes to call it. Here is one possibility based on our experimentation with hand-written parsing code in the Allsorts implementation.

Primitive Types
============

Primitive binary types map directly to the appropriate Rust type:
```
host(U8) = u8
host(U16Be) = u16
host(U16Le) = u16
host(U32Be) = u32
host(U32Le) = u32
```

If Types
=======

Rust does not support type-level conditional expressions so they are mapped to an enum that represents both possible branches, checked at runtime:
```
host(if _Cond then Type1 else Type2) = Either<host(Type1), host(Type2)>
```

Match Types
==========

Match expressions can be desugared to nested if expressions but these will map to nested enums that are awkward to use and inefficient to store. A better approach is to map each match expression to a custom enum type as in this example:
```
host(match _Expr with _Val1 => Type1, _Val2 => Type2, otherwise Type3) = CustomEnum

enum CustomEnum {
    Branch0(host(Type1)),
    Branch1(host(Type2)),
    Branch2(host(Type3)),
}
```
We could extend this compilation scheme with directives to specify suitable names for the custom enum type and its branches to improve the readability of the generated API.

Array Types
==========

Arrays map to Vecs where the length is tracked at runtime and passed as an argument when parsing:
```
host(Array(_Len, Type)) = Vec<host(Type)>
  where ReadBinary<Vec<T>> has Args: usize
```

Function Types
============

```
double : int -> int = \n . n + n

smart_int : bool -> Format = \b . if b then U16Be else U32Be

if_then_else : bool -> Format -> Format -> Format = \b . \t1 . \t2 . if b then t1 else t2

octet : Format -> Format = \t . Array 8 t

wut : (Format -> Format) -> bool -> Format = \f . \b . if b then f U16Be else f U32Be

wut2 : (Format -> Format) -> bool -> Format -> Format = \f . \b . \t . if b then f t else f (f t)
```

```
// double is erased as it does not return a type

// argument is erased, result is a type
type SmartInt = Either<u16, u32>;

trait TypeFunc<T> {
    type ResultType;
}

type TypeCall<S, T> = <S as TypeFunc<T>>::ResultType;

enum Octet {}

[...]

PR #194, opened 2019-10-31
Finish compilation of type-level if expressions

### LINKED ISSUE #170: Boolean eliminators
We'll want to be able to check/compile something like this:

```
test = if true { true } else { false };
Test = if true { F64 } else { F32 };
```

I'm guessing the Rust compiler might emit different variants for each arm of the `if`. Eg.

```rust
enum Test {
    True(F64),
    False(F32),
}
```

brendanzab:
This (finally) completes the compilation of type-level if expressions! If a type-level if expression is detected in a type alias, we compile that type alias into a struct that allows us to capture the type-level computation during parse time.

Closes #170.
Working towards #190.


brendanzab:
Still need to add some rust parse tests before I declare this 'done'! 😅

crates/ddl/src/core/compile/rust/mod.rs
```diff
@@ -142,20 +143,50 @@ fn compile_alias(
...
+                            name,
+                            fields: vec![rust::TypeField {
+                                doc: Arc::new([]),
+                                name: "inner".to_owned(),
+                                format_ty: ty,
+                                host_ty,
+                                by_ref: !is_copy,
+                            }],
+                        })),
+                    ),
```
brendanzab: This may break once we start composing more complicated expressions together. Ie. we could end up with an if expression in the argument to another type alias once we add functions. I have some ideas how to solve this, which would involve propagating the fact that we've seen an 'arbitrarily complicated' if expression upward (when compiling terms).

This feels a tad ad-hoc though, and might be more reasonably solved if we knew the phase that terms could be evaluated in. Then we'd either be able to partially evaluate (if the dependencies of the term were known at compile time) or convert it to a struct (if we were relying on parse-time information). I'm cool with doing the ad-hoc solution for now though.

PR #195, opened 2019-11-12
Consolidate external item handling

brendanzab:
As I've been trying to make headway on a number of issues I've been running up against the pain of not having a nice way to talk about external definitions. This involves lots of copying of the same information across files (eg. the elaborator, delaborator, and grammar), and results in me often forgetting to keep these in sync! This will become more important once we start to add external function definitions.


brendanzab:
This is now kind of handled using globals.

PR #197, opened 2019-11-15
Implement simple match expressions

brendanzab:
This implements simple match expressions for integers. I still need to implement matching on constants, but this can be added later.


brendanzab:
This only supports pattern matching on terms for now - I'm going to leave type-level pattern matches to another pull request!

PR #198, opened 2019-11-21
Update and fix HarfBuzz links

spl:
* Update the HarfBuzz links to use the latest repository. Since GitHub redirects the old <https://github.com/behdad/harfbuzz> to the new <https://github.com/harfbuzz/harfbuzz>, this is merely a nice-to-have.
* Fix the broken “a handy set of types and macros” link. The `hb-open-type-private.hh` file was split into multiple files and no longer exists. I've gleaned that the relevant change occurred in https://github.com/harfbuzz/harfbuzz/commit/92b1e025c639d006f55400bf68fc23bdeaa1c716. 🚧**Please verify that the new link points to the intended code.**

### DIFF
--- a/docs/specification/references.md
+++ b/docs/specification/references.md
@@ -81,7 +81,7 @@ structures, such as file formats, network protocols or bitstreams.
   the Open Type Font specification. Shows promise as nice way to describe
   binary formats, but JS is not the most flexible host language.
 - [HarfBuzz](https://github.com/harfbuzz/harfbuzz) defines [a handy set of types
-  and macros](https://github.com/harfbuzz/harfbuzz/blob/35218c488c3966aa6d459ec5a007a2b43208e97c/src/hb-machinery.hh#L40-L92)
+  and macros](https://github.com/harfbuzz/harfbuzz/blob/35218c488c3966aa6d459ec5a007a2b43208e97c/src/hb-machinery.hh)
   that allows binary data to be cast in-place into declaratively descibed Open
   Type Font tables. Alas, it is in C++, and the API seems extremely prone to
   human-error. A Rust API would be much more robust in regards to upholding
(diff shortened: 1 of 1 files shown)


brendanzab:
Thanks a bunch! Perhaps we can just remove the link then if it's been split up.

spl:
> Perhaps we can just remove the link then if it's been split up.

I tried to guess the section of code referenced by the documentation. If it's wrong, it's easy enough to change. The formal break-up appears to be at https://github.com/harfbuzz/harfbuzz/commit/c77ae40852aabb1698da9e8d488f75bae2f4e734.

An alternative option would be to point to `hb-open-type-private.hh` before the break-up:

<https://github.com/harfbuzz/harfbuzz/blob/6b11fea99716e22522b563d919ef7c72fe9a3c45/src/hb-open-type-private.hh>

brendanzab:
I think just linking to the entire file is good enough - I wanted to include the macros that dealt with sized arrays and structs etc.

docs/specification/references.md
```diff
@@ -80,8 +80,8 @@ structures, such as file formats, network protocols or bitstreams.
   EDSL. Used in [fontkit](https://github.com/devongovett/fontkit) for describing
   the Open Type Font specification. Shows promise as nice way to describe
   binary formats, but JS is not the most flexible host language.
-- [Harfbuzz](https://github.com/behdad/harfbuzz) defines [a handy set of types
-  and macros](https://github.com/behdad/harfbuzz/blob/master/src/hb-open-type-private.hh)
+- [HarfBuzz](https://github.com/harfbuzz/harfbuzz) defines [a handy set of types
+  and macros](https://github.com/harfbuzz/harfbuzz/blob/35218c488c3966aa6d459ec5a007a2b43208e97c/src/hb-machinery.hh#L40-L92)
```
brendanzab:
```suggestion
  and macros](https://github.com/harfbuzz/harfbuzz/blob/35218c488c3966aa6d459ec5a007a2b43208e97c/src/hb-machinery.hh)
```

brendanzab:
Thanks!

PR #212, opened 2020-03-30
Update dependencies

brendanzab:


brendanzab:
Thanks for the review! Yeah that might be an idea… but yeah, I think there are going to be more changes here - like I'll probably remove the `codespan` dependency and just depend on `codespan-reporting`. I'm also planning to make intermediate diagnostic types to make adding new diagnostics easier. So there'll be more updates to come!

PR #215, opened 2020-05-13
Implement a language specification using Makam

brendanzab:
This PR uses the [Makam](https://github.com/astampoulis/makam/) metalanguage (by @astampoulis) to begin implementing a language specification for the DDL. It is an evolution [of this gist that I've been working on](https://gist.github.com/brendanzab/7a331adae167ef6399c6b762aa943abd).

Using a language like Makam is really nice because it's not _so_ dynamic like Prolog is, but it is lightweight enough to sketch out our design without necessarily having to commit to proofs just yet! We can also write unit tests to see if our rules work as expected.

I also really like that the rules are just made up of λProlog-style clauses and datatypes. This means I don't have to constantly come up with new notations for different things, which can become rather challenging in a large system! The module support, while rudimentary, is also helpful when you want to transformations between different source and target languages. In traditional  notations you often have to use different letters for this, or sometimes [even different colours or formatting](https://www.williamjbowman.com/resources/wjb-dissertation.pdf)!

Ultimately it would be really neat if we could produce some kind of nicely formatted documentation from this as part of a [docusaurus](https://v2.docusaurus.io/) site, but that is a way off. For now it's been incredibly useful as a prototyping tool!


brendanzab:
One question I have is that I've made some simplifications to how universes and formats work in this spec. Should update the implementation in this PR, or do that separately? I also might want to delete the old spec in `docs/`.

spec/src/core/typing/declarative.makam
```diff
@@ -0,0 +1,247 @@
...
+    extend : context -> term -> context.
+
+
+    is_context : context -> prop.
+    is_type : context -> term -> prop.
+    is_elem : context -> term -> term -> prop.
+    %  is_subst : context -> subst -> context -> prop.
+    is_equal_type : context -> term -> term -> prop.
+    is_equal_elem : context -> term -> term -> term -> prop.
+    %  is_equal_subst : context -> subst -> subst -> context -> prop.
```
brendanzab, line 32: I'm actually rather proud of the names of these predicates! 😊

spec/src/core/init.makam
```diff
@@ -0,0 +1,32 @@
+% Core language
+%
+% The core language is an extension of Martin-Löf type theory, with some builtin
+% data types, like booleans, integers, and arrays, and support for defining
+% binary format descriptions. We skip identity types and η-rules for now.
+%
+% In making this language, we owe much of our inspiration to the work on PADS/ML
+% and the data description calculus (DDC) in ["The Next 700 Data Description Languages"].
+% In contrast to DDC, we have decided to 'embed' format descriptions within a
+% dependendently typed language rather than splitting the language up into
```
mikeday, line 10: dependendently

brendanzab: oh woopsie haha

spec/src/core/semantics.makam
```diff
@@ -0,0 +1,413 @@
...
+        format_elim First First'.
+
+
+    % Readback of values into terms in normal form
+
+    readback : int -> neutral -> term -> prop.
+    readback : int -> value -> term -> prop.
+
+    % FIXME: Makam's type unification breaks hear for some reason - it
+    % thinks that we want to definine `int -> value -> term -> prop.`
```
astampoulis, line 241: The issue here is that both `readback` and `local` are overloaded and constant resolution fails as a result. Type-based overloading in Makam does not support arbitrary backtracking, but has a few heuristics that unfortunately are not enough here. One option is to avoid defining `readback : int -> value -> term -> prop` until later, if you don't need to mix the rules of the two predicates.

spec/src/init.makam
```diff
@@ -0,0 +1,79 @@
+% A specification of the data description language.
```
astampoulis, line 1: For files like these, you might prefer to use the literate Markdown-based syntax for Makam. You would have to name the file `init.md`, uncomment the documentation, and enclose actual makam code in \`\`\`makam-code blocks.
You would then have to change the `repl` command to `makam src/init.md -`.

spec/src/core/semantics.makam
brendanzab: Ahhh right! Didn't realise it didn't do backtracking. 

Yea I can't really re-order because they are mutually recursive. This is where I kind of wish for languages that used names for 'presentation', but had some kind of unique identifier under the hood. So inputting an identifier would be more like a 'search-disambiguate' operation in your editor. Anyway, that's straying a bit off topic here! 😅

spec/src/core/semantics.makam
```diff
@@ -0,0 +1,407 @@
+% Semantics
+%
+% This section defines an operational semantics for language, using
+% normalization-by-evaluation for performance reasons.
+
+%extend semantics.
+
+    %use functional.
```
brendanzab, line 8: Hmm, this is failing with:
```
 yarn run v1.22.4
$ makam test/init --run-tests

	Makam, version 0.7.35

error Command failed with exit code 1.

!! Error in file ./test/../src/core/semantics.makam:8:5-21:
   File functional.makam not found (searched: [./test/../src/core/functional.makam; /home/runner/work/ddl/ddl/node_modules/makam/functional.makam; /home/runner/work/ddl/ddl/node_modules/makam/stdlib-cache/functional.makam]).

info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
Any ideas @astampoulis?

brendanzab: Ohhh, `%use "stdlib/functional".` seems to work

spec/src/init.makam
brendanzab: I probably would prefer some docs that could be compiled from Makam - with an index and cross-links between symbols. Kind of because this is more something that is intended to be consumed non-linearly, than in the form of a top-down document like is assumed with literate programming.

But yeah, that said, I realise that would be a bunch of work to implement! Doing it with markdown might be a good interim solution because then I could make the specification a subdirectory of a [Docusaurus](https://v2.docusaurus.io/) site, for instance.

spec/src/stdlib.makam
```diff
@@ -0,0 +1,16 @@
...
+    keys : map A B -> set A -> prop.
+    keys [] [].
+    keys ((Key, _) :: Map) (Key :: Keys) :-
+        keys Map Keys.
+
+    map : ((A * B) -> prop) -> map A B -> prop.
+    map P [].
+    map P ((Key, Val) :: Map) :-
+        P (Key, Val),
+        map P Map.
```
brendanzab, line 14: Wondering if these would be handy to have in the standard library @astampoulis?

spec/src/stdlib.makam
```diff
@@ -0,0 +1,22 @@
...
+%extend set.
+
+    map : (A -> prop) -> set A -> prop.
+    map P [].
+    map P (Elem :: Set) :-
+        P Elem,
+        set.remove_if_member Set Elem Set',
+        map P Set.
+
+%end.
```
brendanzab, line 22: Wondering if these extensions would be handy to have in the standard library @astampoulis?

astampoulis: Sure, these would be useful if you feel like opening a PR!

BTW I think the second line is not being used as is, do you mean `map P Set'` in the last line maybe?

spec/src/ast/core/typing/bidirectional.makam
```diff
@@ -109,7 +113,7 @@
     synth_type Context (function_type InputType OutputType) type_type :-
         is_type Context InputType,
         context.eval Context InputType InputType',
-        context.add_param Context InputType' Context'.
+        context.add_param Context InputType' Context',
         is_type Context' OutputType.
```
brendanzab: This was unfortunate! 🤦 🤦

brendanzab: @astampoulis Not sure if Makam could give some warnings on unused variables here? This meant that a bunch of my tests were passing when they should have been failing!

On a related note I've also run into confusing type errors when `:-` is left off by accident, but I think that's pretty hard to avoid, given the λProlog syntax. Thankfully it's less critical because you actually get an error in that case (although it's scary from a user perspective).

brendanzab:
Ok, so I haven't finished work on this, but it has been positive so far! I've moved it into the `experiments/makam-spec` directory for now, and think it might be a good idea to merge this branch. It's become a bit ungainly with so many commits!

PR #226, opened 2020-09-07
Use actions-rs/install for fetching mdbook

brendanzab:


brendanzab:
Hmm, this seems just as slow as before. Seems like the actions-rs folks just have a [big list of crates that they build every night](https://github.com/actions-rs/tool-cache/blob/16cf740aa5bb66a2df71d867e3d77bf67ea5a6d1/.github/workflows/build.yml#L22-L40).

PR #229, opened 2020-09-10
Rename fathom-rt to fathom-runtime

brendanzab:
The shortened name was a bit confusing when explaining the architecture of Fathom! Hopefully this makes it less confusing!


toothbrush:
I'm always in favour of less confusing naming :clap:

PR #230, opened 2020-09-10
Remove Rust backend for now

brendanzab:
The current Rust compiler has become unmaintainable and more thought needs to be put in before progressing. Shedding the compiler in the interim will let us focus our work on the front-end in the short term.


brendanzab:
The test harness for the binary parser tests is still kind of horiffic - calling out to `cargo` from within `cargo` - but I thought it better to preserve the binary interpreter tests that I _do_ have, rather than removing them outright. We'll have to revisit how we do this another time.

PR #234, opened 2020-09-22
Fix a few documentation typos

adrianwong:
These were discovered after a quick scan-through of the docs.


brendanzab:
Thanks for this!

PR #235, opened 2020-09-23
Use Logos for lexing surface language, core languague, and literals

brendanzab:
[Logos](https://github.com/maciejhirsz/logos) has improved greatly since I created the current lexer for Fathom. Taking advantage of it should mean that it's a bit easier to change and extend Fathom going into the future. I've also reworked my implementation of literal parsing, hopefully for the better! 😅


mikeday (approved):
Logos looks pretty handy!

PR #238, opened 2020-10-07
First cut of laying out options

toothbrush:
I've grabbed the options as suggested in #239 by @mikeday, and put them into this simple configuration.  While the flags work pretty much as described in that issue, the output of `--help` leaves a lot to be desired and i'm not sure that [`structopt`](https://github.com/TeXitoi/structopt) actually can accommodate everything we want.

Things i'd still like to fix:

- Make it clear that `--format` and `--format-file` are mutually exclusive, instead of exploding if a user tries to specify both.
- Sensibly group the format- and output-related options in `--help`.  Currently they're just sorted alphabetically.

I don't want to waste too much time on this, i'll revisit it once this package has managed to integrate with `fathom` a little.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 members = [
     './fathom',
+    './fathom-cli',
     './fathom-runtime',
     './fathom-test',
     './fathom-test-util',
--- a/fathom-cli/Cargo.toml
+++ b/fathom-cli/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "fathom-cli"
+version = "0.1.0"
+authors = ["YesLogic Pty. Ltd. <info@yeslogic.com>"]
+edition = "2018"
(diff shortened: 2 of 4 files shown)


wezm (approved):
Looks good to me.

One note: [clap](https://github.com/clap-rs/clap) upon which `structopt` is built is (hopefully) in the latter stages of merging in the `structopt` functionality. It might be worth considering building the CLI on top of the clap 3.0 beta with the assumption that by the time fathom is ready so will clap 3.

toothbrush:
> Looks good to me.
> 
> One note: [clap](https://github.com/clap-rs/clap) upon which `structopt` is built is (hopefully) in the latter stages of merging in the `structopt` functionality. It might be worth considering building the CLI on top of the clap 3.0 beta with the assumption that by the time fathom is ready so will clap 3.

Ah, that's a really helpful remark.  Thanks.  For now i think i'll just keep this as it's "good enough" for testing, but i'll avoid investing too much time into this approach then.

PR #245, opened 2020-10-15
Implement struct types

### LINKED ISSUE #241: Non-dependent struct formats
For example:

```
struct Point : Format {
    x : S32Be,
    y : S32Be,
}
```

brendanzab:
I had originally hard-coded the `struct` syntax to define struct formats. This changes it so that we now elaborate to either a struct type or a struct format in the core language depending on an annotation. Not supplying an annotation is an error.

Closes #241 and #242 .


fathom/src/pass/surface_to_core.rs
```diff
@@ -144,89 +144,42 @@ impl<'me> Context<'me> {
...
-                                });
-                            }
-                        }
-                    }
+                ItemData::StructType(struct_type) => {
+                    let (_, r#type) = self.synth_type(file_id, &struct_type.type_);
+                    let item_data = match r#type.as_ref() {
+                        Value::Sort(Sort::Type) => self.is_struct_type(file_id, struct_type),
+                        Value::FormatType => self.is_struct_format(file_id, struct_type),
+                        r#type => todo!("expected Type or Format"),
```
brendanzab, line 160: Need a proper diagnostic here.

toothbrush (approved):

PR #246, opened 2020-10-15
Add some aspirational examples

brendanzab:


brendanzab:
I think STL is ready to go - just have questions about how we want to integrate this into the test suite. I'd prefer these to go into some kind of 'format gallery' in a top level directory, perhaps in `examples`, and I'm wary of using the existing testing infrastructure (see #253).

wezm:
This is totally an aside but having a a working EDID codegen path would actually be useful. [I had the need to edit an EDID definition](https://www.wezm.net/v2/posts/2020/linux-amdgpu-pixel-format/) earlier in the year and the only tool to do this was super crashy but do the job in the end.

brendanzab:
Oooh cool! Nice to know there is some need for this. I think the main things we would be missing for EDID is parsing sub-byte stuff nicely, doing computed fields, and repeating until the end of the file. I think some of that could be bodged on the Rust side of things, but it would certainly be nice to have it in the DDL!

Let me know if you come across any other formats that might be interesting to sketch out!

Issue #244, opened 2020-10-09
Dependent struct formats and struct types


brendanzab:
This will allow dependent struct formats to have field dependencies, which correspond to how structs are parsed in a binary format. For example:

```
struct Points : Format {
    len : U32Be,
    data : FormatArray len { x : U32Be, y U32Be },
}
```

Note that `len` is bound as `len : Repr U32Be` in the typing context. 

Unlike DDC we aim to preserve as much dependency information in the host language as possible. To that end `Repr` will interpret dependent struct formats into dependent struct types (as opposed to a non-dependent structs). 

We need to be careful of maintaining dependencies when looking up fields in these records. For example:

```
struct MyArray {
    len : Int,
    data : Array len Point,
}

get_data (array : MyArray) =
    array.data; // type of body is `Array array.len Point`
```

This will require us to implement local environments. When doing so we'll also have to update the `Bool` and `Int` eliminators to capture their current environment (lest we introduce weird scoping behaviour!)

toothbrush:
Here's a note: https://github.com/yeslogic/fathom/pull/248/commits/ec7e6829e80eb12d2b4cd8500fae4e35f433cf44#r508133011

PR #248, opened 2020-10-19
Implement struct terms and eliminations

### LINKED ISSUE #247: Struct terms and eliminations
For example, given the following `Point` struct:

```
struct Point : Type {
    x : Int,
    y : Int,
}
```

We'll want to be able to construct a term of that type, and project on its fields:

```
origin : Point = struct {
    x = 0,
    y = 0,
};
origin_x = origin.x;
origin_y = origin.y;
```

Add ability to construct terms of the type `struct` that we are able to introduce, as well as provide access to fields.

brendanzab:
This implements struct terms and eliminations. For example, we can now write:

```
struct Point : Type {
    x : Int,
    y : Int,
}

const origin : Point = struct {
    x = 0,
    y = 0,
};

const origin_x = origin.x;
const origin_y = origin.y;
```

Closes #247.


book/src/reference/structs.md
```diff
@@ -0,0 +1,80 @@
...
+```fathom
+point : Point = struct {
+    x = 23,
+    y = 42,
+};
+
+point.x // normalizes to `23`
+```
+
+When synthesizing the type of field lookups we need to be careful about data dependencies:
```
toothbrush, line 75: Re: not forgetting the context of `MyArray` which contains `len` and `data`, they need to remain correctly lexically scoped so that when you try to use `array.data` down below, you're able to correctly infer/check that it has type `Array len Point`.

Is it worth creating an issue about this so we don't forget?

brendanzab: Yeah, we could add it to #244 perhaps?

fathom/src/lang/core/typing.rs
```diff
@@ -155,6 +154,42 @@ impl<'me> Context<'me> {
...
+                        if !seen_field_names.insert(field.name.clone()) {
+                            self.push_message(CoreTypingMessage::FieldRedeclaration {
+                                file_id,
+                                field_name: field.name.clone(),
+                                record_range: item.range(),
+                            });
+                        }
+                    }
+
+                    // FIXME: Avoid shadowing builtin definitions
```
toothbrush, line 176: From discussion: is this `FIXME` still a thing?

fathom/src/pass/surface_to_core.rs
```diff
@@ -144,89 +144,65 @@ impl<'me> Context<'me> {
...
+                            file_id,
+                            name: struct_type.name.data.clone(),
+                            name_range: struct_type.name.range(),
+                        });
+                        continue;
                     }
+                    Some(r#type) => {
+                        let (core_type, _) = self.is_type(file_id, &r#type);
+                        let r#type = self.eval(&core_type);
+                        let item_data = match self.eval(&core_type).as_ref() {
```
toothbrush, line 167:
```
                         let item_data = match self.eval(&core_type).as_ref() { 
```

Let's not redo evaluation of the `core_type`.

tests/input/fail_duplicate_definitions.fathom
```diff
@@ -1,3 +1,4 @@
-struct Empty {}
-struct Empty {} //~ error: the name `Empty` is defined multiple times
-struct Empty {} //~ error: the name `Empty` is defined multiple times
+struct Empty : Format {}
+struct Empty : Format {} //~ error: the name `Empty` is defined multiple times
+struct Empty : Format {} //~ error: the name `Empty` is defined multiple times
+Empty = 1 : Int; //~ error: the name `Empty` is defined multiple times
```
toothbrush, line 4: Maybe it'd be helpful to have the user write `const Empty = 1 : Int;` when introducing this binding, otherwise it reads a bit like defining the type of `Empty` as `Format` and subsequently assigning a value to it.

brendanzab: Made an issue for this here: #250

toothbrush: Cool, thanks – i like being organised with cards we can work on :nerd_face:

toothbrush (approved):
Apologies for the delay, but i finally had a look through this.  From what i can see it looks very reasonable, thanks for cleaning up the error handling i didn't manage to do.

I don't feel like i can really speak to the specifics, but if you're waiting for my blessing, i'd say merge away! :+1: :shipit:

toothbrush:
Merging after discussion!

PR #251, opened 2020-10-21
Implement `repr` type operator

### LINKED ISSUE #243: Built-in 'Repr' function
This converts binary format descriptions into their host representation types. This will be important once we want to implement dependent record formats.

Initially this would correspond with the following implementation: 

```
Repr : Format -> Type
Repr U8     = Int
Repr U16Le  = Int
Repr U16Be  = Int
Repr U32Le  = Int
Repr U32Be  = Int
Repr U64Le  = Int
Repr U64Be  = Int
Repr S8     = Int
Repr S16Le  = Int
Repr S16Be  = Int
Repr S32Le  = Int
Repr S32Be  = Int
Repr S64Le  = Int
Repr S64Be  = Int
Repr F32Le  = F32
Repr F32Be  = F32
Repr F64Le  = F64
Repr F64Be  = F64
Repr { l₀ : t₀, ... lₙ : tₙ } =
    { l₀ : Repr t₀, ... lₙ : Repr tₙ }
```

For example:

```
struct Point : Format {
    x : S32Be,
    y : S32Be,
}

origin : Repr Point = struct {
    x = 0,
    y = 0,
};
```

brendanzab:
This implements the `repr` type operator. This converts `Format` types into `Types`, which will become important once we start implementing data-dependent format descriptions (for example dependent struct types).

Some examples are:

```
repr U32Be                  // Int
repr U32Le                  // Int
repr (FormatArray len Elem) // Array len (repr Elem)
```

I've implemented this as a keyword, not as a global function, as we'll need some special typing rules around handling format items (ie. when they are eliminated and introduced).

This models the following operator from [DDC](http://portal.acm.org/citation.cfm?doid=1111037.1111039):

Screen Shot 2020-09-16 at 4 25 44 pm

And the one found in [The Power of Pi](https://dl.acm.org/doi/abs/10.1145/1411204.1411213):

Screen Shot 2020-10-21 at 6 46 40 pm
Screen Shot 2020-10-01 at 11 25 53 am

Closes #243.


fathom/src/lang/core/semantics.rs
```diff
@@ -54,7 +57,8 @@ pub enum Head {
     Error,
 }
 
-/// An eliminator that cannot be reduced further is 'stuck' on some [`Head`].
+/// An eliminator that cannot be reduced further due to being stuck on some
```
toothbrush, line 60: I think there's a typo here?

toothbrush (approved):
I can't say i fully understand everything, but i'm excited to have a `repr` in our main branch, so how about we go ahead and revise later if necessary?

fathom/src/lang/core/semantics.rs
brendanzab: I've updated the docs in a commit!

fathom/src/lang/core/semantics.rs
```diff
@@ -54,7 +57,8 @@ pub enum Head {
...
+/// 'stuck' on some [`Head`].
```
toothbrush: I think there's just a bit of brokenness in the wording here.

```suggestion
/// An eliminator that cannot be reduced further due to being
/// 'stuck' on some [`Head`].
```

Maybe like that?

brendanzab: Oh woops hah

brendanzab: I think that was fixed in the later documentation commit

toothbrush: I think i might have somehow been seeing an outdated version – perhaps this can be ignored :sweat_smile:

brendanzab: Yeah, I changed the wording in cbfa5844227bc059a2d176a350709359f065c06

brendanzab:
As discussed in private chats, the name, `repr` is not the greatest! It would be nice to come up with something better later on!

Issue #254, opened 2020-10-29
Array terms


brendanzab:
We should be able to introduce terms of type `Array len T`. For example:

```
good_array : Array 3 Int = [ 0, 0, 0 ];
bad_array : Array 3 Int = [ 0, 0 ]; //~ error: mismatched array length
```

### Concrete Syntax

```
atomic-term : surface-term ::=
    | ...
    | '[' (<terms : term> ',' )* <last-term : term?> ']' => array-term(terms last-term)
```

### Surface Language

```
surface-term ::=
    | ...
    | array-term(surface-term*)
```

### Core Language

```
core-term ::=
    | ...
    | array-term(core-term*)
```

### Semantics

```
value ::=
    | ...
    | array-term(value*)
```

```
eval(array-term(core-term₀ ... core-termₙ)) =
    array-term(eval(core-term₀) ... eval(core-termₙ))
```

```
readback(array-term(core-value₀ ... core-valueₙ)) =
    array-term(readback(core-value)₀ ... readback(core-value)ₙ)
```

### Typing rules

```
rule check-type/array-intro
    check-type(context, array-term(surface-term₀ ... surface-termₙ), Array n elem-type)
when
    check-type(context, surface-term₀, elem-type)
    ...
    check-type(context, surface-termₙ, elem-type)
```

### Elaboration rules

```
rule check-type/array-intro
    check-type(
        context,
        array-term(surface-term₀ ... surface-termₙ),
        array-term(core-term₀ ... core-termₙ),
        Array n elem-type,
    )
when
    check-type(context, surface-term₀, core-term₀, elem-type)
    ...
    check-type(context, surface-termₙ, core-termₙ, elem-type)
```

PR #256, opened 2020-11-04
Allow dependent fields in struct formats and types

### LINKED ISSUE #244: Dependent struct formats and struct types

brendanzab:
This allows for dependencies between fields in struct formats and types. This allows us to express more interesting, data-dependent struct formats, for example:

```
struct MyData : Format {
    count : U32Be,
    data : FormatArray count U32Be,
}
```

Closes #244.


brendanzab:
I'll need to rebase this on top of the work on #247. I also need to implement some binary parser tests for the dependent formats.

brendanzab:
I've rebased this, but I need to fix the elaboration and typing of struct terms and eliminations. I also think it might be useful to implement #254 before hand in order to fully test this.

PR #260, opened 2020-11-10
Add const keyword, rename aliases to constants

### LINKED ISSUE #250: Rename aliases to 'constants' and use a keyword
I have been calling top-level definitions 'aliases' to distinguish them from struct and enum definitions. The name was inspired by 'type aliases' in languages like Rust:

```
magic : Int = 83;
add_one (n : Int) : Int = n + 1;

Tag : Type = F32Be;
Array3 (Elem : Type) : Type = Array 3 Elem;
```

Maybe it would be less confusing to call them 'constants' and use the `const` keyword?

```
const magic : Int = 83;
const add_one (n : Int) : Int = n + 1;

const Tag : Type = F32Be;
const Array3 (Elem : Type) : Type = Array 3 Elem;
```

toothbrush:
This PR introduces a `const` keyword into the core and surface language, the intention being that what was previously called an "alias" will now be referred to as a "constant", as in a constant let-binding.

We have updated the unit tests to reflect this new syntax – unfortunately because most lines are touched Git makes a mess of the diff, but the changes are very minor.

Additionally, we have made some refactoring changes:

- Previously we referred to the collection of integers, floats, etc. as "constants", but we've renamed them to "primitives".  That feels a bit more natural, and frees up the identifier `Constant` for use in the sense of a term bound to a name.
- Performed the rename from `Alias` to `Constant` in code & documentation.

Closes #250


brendanzab:
Awesome! Thinking we should change the `alias` keyword in the core language to `const` as well, and update the ASTs/identifiers/docs to suit.

toothbrush:
> Awesome! Thinking we should change the `alias` keyword in the core language to `const` as well, and update the ASTs/identifiers/docs to suit.

Just to clarify, it looks to me like the core language didn't have an `alias` keyword – i've introduced the `const` keyword, though.

fathom/src/pass/surface_to_doc.rs
```diff
@@ -82,43 +82,47 @@ struct ItemMeta {
...
             None => writeln!(
                 writer,
                 r##"          <a href="#{id}">{name}</a>"##,
                 id = id,
-                name = alias.name.data,
+                name = constant.name.data,
             )?,
             Some(r#type) => writeln!(
                 writer,
                 r##"          <a href="#{id}">{name}</a> : {type_}"##,
```
brendanzab, line 106: We should probably add a `const` keyword here:
```suggestion
                r##"          const <a href="#{id}">{name}</a> : {type_}"##,
```
Note that this will require rebuilding the snapshots.

book/src/specification/textual-representation/concrete-syntax.md
```diff
@@ -17,22 +17,23 @@ process.
...
 
-### Alias definitions
+### Constant definitions
 
-Alias definitions are used to give names to terms that can be later used in
+Constant definitions are used to give names to terms that can be later used in
 other places in the binary description. For example:
 
 ```fathom
 Byte = U8;
```
brendanzab, line 26: There are some constant definitions like this in the book that we'll have to update. Sadly the book is not covered by the tests yet.

fathom/src/pass/surface_to_doc.rs
toothbrush: Ah yes, good catch, thanks!

book/src/specification/textual-representation/concrete-syntax.md
toothbrush: Ah yes, another good catch – i remember you mentioning this and being unsure of when it'd become relevant!

book/src/specification/textual-representation/concrete-syntax.md
```diff
@@ -68,7 +69,7 @@ Modules are lists of zero-or-more definitions. Definitions within a module must
 > <sub>Grammar:</sub>
 >
 > _item_ ::=\
-> &emsp;|&ensp;_alias-type-definition_\
+> &emsp;|&ensp;_constant-type-definition_\
```
toothbrush, line 72: It feels like `constant-type-definition` is unbound -- should that be `constant-definition`?

brendanzab: Yup, agreed

book/src/specification/textual-representation/concrete-syntax.md
```diff
@@ -17,22 +17,23 @@ process.
...
-### Alias definitions
+### Constant definitions
 
-Alias definitions are used to give names to terms that can be later used in
+Constant definitions are used to give names to terms that can be later used in
 other places in the binary description. For example:
 
 ```fathom
-Byte = U8;
+const Byte = U8;
```
brendanzab, line 26: There are more examples than just this one in the book ;)

toothbrush: Oops, i just grepped for "constant" or "alias" :stuck_out_tongue:

toothbrush: I had a look at each Markdown file in the `book` directory, i think it's done now?

brendanzab: Looks great!

brendanzab (approved):
Looks good! Nice work!

PR #263, opened 2020-11-16
Tree wrangling cleanups

brendanzab:
Some assorted cleanups related to wrangling ASTs.

- adds a `Range` type that can be copied
- adds `Value::{try_global, try_item}` shortcut methods to clean up some pattern matching

This is implemented on-top of #248


toothbrush (approved):
Cool, i like it.  Simplifies the matching blocks a lot! :desert_island:

PR #264, opened 2020-11-18
Github Actions disables add_path: fix!

toothbrush:
We should adapt now that Github has [deprecated the old way of setting environment variables](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/).

Since i'm a pedant, i've also added quoting around shell arguments that could cause word-splitting.

Also killing a warning from Github Actions by explicitly setting Ubuntu version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   spec:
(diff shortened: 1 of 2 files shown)


brendanzab (approved):
Thanks for this!

PR #265, opened 2020-11-20
Implement array terms

brendanzab:
This lets us construct inhabitants of `Array len Elem`:

```
[1, 2, 3] : Array 3 Int
```

These are called ‘sequence terms’ in the surface language as they might be overloaded for other purposes in the future.

Closes #254.

PR #266, opened 2020-11-23
Return values from binary reader

brendanzab:
Thanks to the addition of array terms (#265) and struct terms (#248) we are now able to represent all format representations in the host language. This lets us remove `lang::core::binary::Term`, and return values from the interpreted binary reader instead, making it easier to implement dependent parsing in the future (see #244).


toothbrush (approved):
This seems like a good clean-up! :broom: :sparkles:

PR #267, opened 2020-11-25
Allow dependent fields in struct formats and types

### LINKED ISSUE #244: Dependent struct formats and struct types

toothbrush:
Main branch renamed to ... `main`.  Reopening, but this is morally still #256 (there is relevant discussion there). 

This allows for dependencies between fields in struct formats and types. This allows us to express more interesting, data-dependent struct formats, for example:

```
struct MyData : Format {
    count : U32Be,
    data : FormatArray count U32Be,
}
```

Closes #244.


.github/workflows/ci.yml
```diff
@@ -3,11 +3,11 @@ name: ci
 on:
   push:
     branches:
-      - master
+      - main
```
toothbrush, line 6: Oops, totally missed that! :facepalm:

fathom/src/lang/core.rs
```diff
@@ -276,3 +280,145 @@ impl Default for Globals {
         Globals::new(entries)
     }
 }
+
+/// A [de Bruijn index][de-bruijn-index] in the local environment.
+///
+/// De bruijn indices describe an occurrence of a variable in terms of the
```
toothbrush, line 286:
```suggestion
/// De Bruijn indices describe an occurrence of a variable in terms of the
```

Nitpick!

fathom/src/lang/core.rs
```diff
@@ -276,3 +280,145 @@ impl Default for Globals {
...
+/// This describes an occurrence of a variable by counting the binders from
+/// the top of the term towards until the occurrence is reached.
+/// For example:
+///
+/// | Representation    | Example (S combinator)  |
+/// | ----------------- | ----------------------- |
+/// | Named             | `λx. λy. λz. x z (y z)` |
+/// | De Bruijn levels  | `λ_. λ_. λ_. 0 2 (1 2)` |
+///
+/// Levels are used in [values][semantics::Value] because they are not context-
```
toothbrush, line 325: Interesting!  I was unaware of this technique :bulb:

toothbrush:
So far i've mainly digested the test case fixtures, but that looks really good!

fathom/src/lang/core.rs
brendanzab: Oooh thanks!

fathom/src/lang/core.rs
brendanzab: Yeah it's a super nifty way of doing dependent type checking

wezm: > This describes an occurrence of a variable by counting the binders from the top of the term towards until the occurrence is reached.

Is there a word missing after towards?

brendanzab: Good catch! Rewording this to:

> This describes an occurrence of a variable by counting the binders inwards from the top of the term until the occurrence is reached.

PR #268, opened 2020-11-25
Add some aspirational examples

toothbrush:
Main branch renamed to `main` :hand: :fr: Reopening, but this is morally still #246 (see discussion there).


brendanzab:
Think these should be ready to merge! 🤩 The examples aren't yet finished, but they pass the type checker at least!

PR #273, opened 2020-12-03
API for test suite and CLI

### LINKED ISSUE #261: Create top-level compiler API
Currently the test suite reaches into compiler internals.  We should extract that, to unblock creating a CLI.  The test suite & CLI can then use that API.

toothbrush:
The idea here is to pull the guts of the compiler, which are fairly intertwined with the test suite at the moment, into a "compiler driver" module to be created in the `fathom` crate.  Eventually, the hope is that this closes #261.

For now, :warning: this is still extremely messy, and will need a lot of TLC :warning:

### DIFF
--- a/fathom-cli/src/main.rs
+++ b/fathom-cli/src/main.rs
@@ -1,3 +1,4 @@
+use fathom::driver;
 use structopt::StructOpt;
 
 mod cli;
@@ -7,4 +8,19 @@ fn main() {
     println!("Hello, world!");
 
     println!("debug: {:?}", args);
+
+    if args.files.len() == 0 {
+        // error, please provide an input file
+    }
+
+    if args.files.len() != 1 {
+        // note that we only will read one input file.
+    }
+
+    let file = args.files.first().unwrap();
+
+    driver::compiler::run_compiler(
+        &*file.clone().into_os_string().into_string().unwrap(),
+        &*file,
+    )
 }
--- a/fathom-cli/src/main.rs
+++ b/fathom-cli/src/main.rs
@@ -1,26 +1,46 @@
 use fathom::driver;
 use structopt::StructOpt;
 
+use fathom::lang::core;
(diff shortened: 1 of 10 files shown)


fathom-cli/src/main.rs
```diff
@@ -1,10 +1,46 @@
...
 use structopt::StructOpt;
 
+use fathom::lang::core;
+use fathom::lang::core::binary;
+use fathom_runtime::ReadScope;
+
 mod cli;
 
-fn main() {
+fn main() -> Result<(), String> {
```
brendanzab, line 10: We could return an `anyhow::Result` here

fathom/src/driver/compiler.rs
```diff
@@ -0,0 +1,164 @@
...
+            // test status output.
+
+            let mut buffer = BufferWriter::stderr(ColorChoice::Auto).buffer();
+            for diagnostic in &found_diagnostics {
+                term::emit(&mut buffer, &self.term_config, files, diagnostic).unwrap();
+            }
+
+            eprintln_indented(4, "", "---- found diagnostics ----");
+            eprintln_indented(4, "| ", &String::from_utf8_lossy(buffer.as_slice()));
+            eprintln!();
```
brendanzab, line 149: Probably don't need this kind of output for the CLI - perhaps the client of the compiler could handle the diagnostic rendering instead? So there's a `.drain_messages` method on the compiler, perhaps.

brendanzab: Stuff like `---- found diagnostics ----` was there for the tests to distinguish it from the regular rust/test harness output.

toothbrush: Fair enough!  Just so i understand, are you suggesting that the compiler simply expose the list of messages, and the CLI handle their formatting?

brendanzab:
This was addressed as part of #283 :)

PR #277, opened 2020-12-16
Bump cargo dependecies

brendanzab:


toothbrush (approved):
Seems legit! Thanks! :+1:

PR #283, opened 2021-02-16
Add a driver API and use it to implement the beginnings of a CLI

### LINKED ISSUE #261: Create top-level compiler API

brendanzab:
It took a surprising amount of work to figure out the boundaries of this! 😅 Hopefully this will make the Fathom codebase more approachable - as it provides an entrypoint for new contributors to see how things fit together. I expect this will change and evolve as we go, but it's a start at least!

The current CLI commands implemented are:

- `fathom data`: for manipulating binary data using Fathom format descriptions
- `fathom compile`: for compiling a Fathom binary format for a given target (TODO)
- `fathom check`: for checking that Fathom binary formats are valid
- `fathom doc`: for generating documentation for Fathom binary formats

I have a bunch of TODOs in the source code showing things we could add/improve in the future.

Closes #261


brendanzab:
Probably should add some CLI tests for this… maybe using something like what is [talked about in the CLI book](https://rust-cli.github.io/book/tutorial/testing.html)?

Issue #281, opened 2021-01-15
Dependent function types


brendanzab:
It would be nice to be able to define dependent function types. For example:

```
fn (A : Type) (data : Array 3 A) -> A
```

PR #288, opened 2021-03-23
Implement parameterized structs

### LINKED ISSUE #255: Parameterised struct formats
It would be nice to have parameters on struct formats. For example:

```
struct MyTable (SubTable : Format) : Format {
    count : U32Be,
    sub_tables : FormatArray count SubTable,
}
```

The parameters should form a _telescope_, allowing for dependencies between parameters:

```
struct Data (len : Int) (Data : FormatArray len U8) : Format {
    data : Data,
}
```

brendanzab:
This PR implements parameterized struct types and struct formats. This allows us to finally start implementing part of OpenType, which is extremely exciting! 

An example of how this looks in practice is:

```rust
struct GenericPoint (Coordinate : Format) : Format {
    x : Coordinate,
    y : Coordinate,
}

struct MyGenericArray (len : Int) (Coordinate : Format) : Format {
    data : FormatArray len (GenericPoint Coordinate),
}
```

As part of this change we also allow the representations of struct formats to be introduced and eliminated. This had been previously documented, but we had somehow missed implementing it.

Closes #255.


brendanzab:
I still need to check the OpenType example in CI, so we should hold off merging until then!

fathom/src/lang/core.rs
```diff
@@ -83,6 +83,8 @@ pub struct StructType {
     pub doc: Arc<[String]>,
     /// Name of this definition.
     pub name: String,
+    /// Parameter telescope.
```
brendanzab, line 86: Ahh, I guess this isn't really a telescope just yet - the parameters can't depend on each other yet!

fathom/src/lang/core.rs
```diff
@@ -83,6 +83,8 @@ pub struct StructType {
...
+    pub params: Arc<[(Located<String>, Arc<Term>)]>,
```
brendanzab, line 87: Ahh, I guess this isn't actually a telescope just yet - the parameters can't depend on each other yet! We'd need to implement dependent function types first (see #281).

fathom/src/lang/core/typing.rs
```diff
@@ -331,25 +355,22 @@ impl<'me> Context<'me> {
...
-                            let r#type =
-                                self.eval_with_locals(&mut type_locals, &field_declaration.type_);
-                            // TODO: stop on errors
-                            self.check_type(&field_definition.term, &r#type);
-                            let value = self.eval(&field_definition.term);
 
-                            type_locals.push(value);
+                field_declarations.for_each_field(
+                    self.globals,
+                    &self.item_definitions.clone(), // FIXME: avoid clone
```
brendanzab, line 361: It's kind of frustrating to have to do these clones of the item definitions - but it's required seeing as we take `self` by mutable reference in the closure, even though we don't mutate `items` at all. Perhaps we could use [`elsa::FrozenMap`](https://docs.rs/elsa/1.4.0/elsa/map/struct.FrozenMap.html) to make this a bit nicer?

mikeday (approved):
Looks good!

PR #293, opened 2021-05-06
Begin to rework integration tests

brendanzab:
This improves the test harness in the following ways:

- [x] uses `libtest-mimic` to dynamically construct a `rustc --test`-style test harness
  - [x] integration tests are now automatically discovered 
  - [x] examples tests are run in parallel, at the same time as the integration tests
- [ ] integration tests now use the CLI rather than reaching into the internals of Fathom
- [ ] the cursed diagnostic lexer has been removed


brendanzab:
Seems like in order to make progress on decoupling the tests from the internals I might want some simple JSON serialization of the compiler diagnostics in order to correlate the line `//~ error: ...` directives with the compiler output, without having to match on the pretty-printed output. This might be a candidate for inclusion in `codespan-reporting`, but it could initially be done in `codespan-cli`. Not sure I want to do this in this PR.

brendanzab:
Haven't got everything finished, but I'm gonna merge this as is, because I think it's an improvement. I was also experimenting with TOML-based syntax for test directives in Pikelet, which would avoid me having to come up with something myself: https://github.com/pikelet-lang/pikelet/pull/258

PR #306, opened 2022-02-22
Implement overlap formats

brendanzab:
This implements a format comprised of a dependent sequence of overlapping formats. This allows later information to affect how past information is parsed, which could be useful for defining the tricky mutual recursion found in the font tables in OpenType.

### DIFF
--- a/fathom/src/core.rs
+++ b/fathom/src/core.rs
@@ -142,6 +142,9 @@ pub enum Term<'arena> {
     ArrayIntro(&'arena [Term<'arena>]),
     /// Record formats, consisting of a list of dependent formats.
     FormatRecord(&'arena [StringId], &'arena [Term<'arena>]),
+    /// Overlap formats, consisting of a list of dependent formats, overlapping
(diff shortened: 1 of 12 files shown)


brendanzab:
An alternative approach to tackling the 'weird mutual recursion' of the OpenType font table formats might be:

```
let table_directory (file_start : Pos) = {
    ...

    // Parse the font table records (with offsets to the font tables)
    table_records <- array16 num_tables table_record,

    // These fields don't consume the stream - they just register offsets to the the font tables
    // `sequence16` takes an array of format descriptions and parses them in turn
    font_tables0 <- sequence16 _ (map16 _ _ _ (font_table0 file_start) table_records),
    font_tables1 <- sequence16 _ (map16 _ _ _ (font_table1 file_start _ font_tables0) table_records),
};
```

It's hard to know which one I prefer though, so might be worth seeing if `overlap` gets any use… we can always remove it later on if it doesn't end up being useful.

PR #307, opened 2022-03-03
Link formats

brendanzab:
The link formats could be simplified to a single `link` format if we allowed for limited arithmetic on positions. For now I just have a bunch of link formats that take a position and an offset:

- `Ref : Type -> Type`
- `link8 : Pos -> U8 -> Format -> Format`
- `link16 : Pos -> U16 -> Format -> Format`
- `link32 : Pos -> U32 -> Format -> Format`
- `link64 : Pos -> U64 -> Format -> Format`

Todo:

- [x] integration tests
- [x] check the open type format works with the new links
- [x] write reference documentation

### DIFF
--- a/fathom/src/core.rs
+++ b/fathom/src/core.rs
@@ -193,6 +193,8 @@ pub enum Prim {
     Array64Type,
     /// Type of stream positions.
     PosType,
+    /// Type of stream references.
+    RefType,
 
     /// Type of format descriptions.
     FormatType,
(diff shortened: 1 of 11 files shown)


fathom/src/core.rs
```diff
@@ -246,6 +248,18 @@ pub enum Prim {
...
+    /// relative to a base position and a 8-bit offset.
+    FormatLink8,
+    /// A format that links to another location in the binary data stream,
+    /// relative to a base position and a 16-bit offset.
+    FormatLink16,
+    /// A format that links to another location in the binary data stream,
+    /// relative to a base position and a 32-bit offset.
+    FormatLink32,
+    /// A format that links to another location in the binary data stream,
+    /// relative to a base position and a 64-bit offset.
```
mikeday, line 261: Unsigned offsets, presumably.

brendanzab: ahh, that's correct!

brendanzab: Hopefully avoiding obvious non-termination issues but idk

mikeday: The only way you could get non-termination is by repeatedly loading the same offset with different types, but you would need an infinite type to make that work. You could come close with something like this:
```
struct Nasty (r: Ref) (n: u32) {
    oops : Link32 r 0 (Nasty r (n + 1))
}
```
Perhaps you could make this worse by evaluating Ackermann's function in the type? :laughing:

brendanzab:
This has exposed a bug in how I handle `Repr`, which I'm currently tracking down!

PR #310, opened 2022-03-10
Add some length limiting formats

brendanzab:
This implements some length limiting formats:

- `limit8 : U8 -> Format -> Format`
- `limit16 : U16 -> Format -> Format`
- `limit32 : U32 -> Format -> Format`
- `limit64 : U64 -> Format -> Format`

These limit the scope that a format is parsed in, from the current position in the binary data, to an exact length, failing if there is not enough room left in the buffer. I'm not sure if this is the best name for them. Maybe something that gives a sense that we are creating a new ‘scope’ would be better.

In order to implement this I went back to an Allsorts-style approach to parsing the binary data, decoupling it from `std::io`. I was working on introducing the format traits from Allsorts as well, which would better show the connection between `Repr` and the rust traits, but I've scoped back that change for now as it was taking me far too long to figure out.


brendanzab:
Moving this along as I figure out a good approach for dereferencing.

fathom/src/core/binary.rs
```diff
@@ -1,17 +1,188 @@
...
+/// -n       ...         0     1     2     3     4     5     6     7     8      ...      m-1   m
+///  │                   │                                                                     │
+///  └─ Buffer::offset ─>└─────────────────────── Buffer::remaining_len ──────────────────────>│
+///  │                   │                                   │                                 │
+///  │                   └── BufferReader::relative_offset ─>│                                 │
+///  │                                                       │                                 │
+///  └─────────────── BufferReader::offset ─────────────────>└── BufferReader::remaining_len ─>│
+/// ```
+#[derive(Copy, Clone)]
+pub struct Buffer<'data> {
```
brendanzab, line 42: Perhaps `SubBuffer` or `BufferScope` would be a better name for this.

fathom/src/main.rs
```diff
@@ -163,8 +164,13 @@ fn main() -> ! {
                 Input::File(path) => driver.read_source_path(&path),
             };
 
-            let mut reader = BufReader::new(std::fs::File::open(binary_path).unwrap()); // TODO: report errors
-            let status = driver.read_format(file_id, &mut reader);
+            // TODO: report errors
+            let mut file = std::fs::File::open(binary_path).unwrap();
+            let mut data = Vec::new();
+            file.read_to_end(&mut data).unwrap();
+            let buffer = fathom::core::binary::Buffer::from(data.as_slice());
```
brendanzab, line 171: Need to deal with these errors better. Maybe I should move the binary file processing into the driver 🤔

brendanzab:
Kaitai seems to have [“Substreams”](https://doc.kaitai.io/user_guide.html#stream) which seems pretty related to this. They deal with relative and absolute positioning which is interesting.

fathom/src/core/binary.rs
```diff
@@ -1,17 +1,188 @@
...
+    /// Offset from the starting position.
+    offset: usize,
```
wezm, line 44: Minor: `offset` kinda feels mildly confusing because at first glance it looks like it would be an offset within `data`. Perhaps `start_pos`, `base_pos`, `base_offset` would be clearer?

fathom/src/core/binary.rs
```diff
@@ -1,17 +1,188 @@
...
+    pub fn len(&self) -> Result<usize, ReadError> {
+        usize::checked_add(self.offset, self.data.len()).ok_or(ReadError::PositionOverflow)
+    }
+
+    /// Remaining number of bytes in the buffer.
+    pub fn remaining_len(&self) -> usize {
+        self.data.len()
+    }
+
+    /// Return a buffer limited to the remaining length of the buffer.
```
wezm, line 70: ?

```suggestion
    /// Return a buffer limited to the supplied length.
```

wezm (approved):
Looks good!

brendanzab:
Thanks Wes! I'll fix those things.

Thinking about it a bit, I think the reason why I'm uncomfortable with the deref behavior (which triggers an error when you deref outside the current buffer) is that you can still register links outside of the current buffer. I dunno why, but I kind of feel that these should be consistent. Ie. either trigger a parse error if there is a link is registered outside the current buffer, or allow derefs to escape.

fathom/src/core/binary.rs
brendanzab: Do you think it would be weird to change the name of `BufferReader::offset` to something similar?

brendanzab: Or maybe we could call it `BufferReader::absolute_offset` or something 🤔

brendanzab:
I guess if we also go with the idea that links are scoped to the current buffer, then maybe the should also be parsed within the bounds of the buffer they were created in. Which also might interact weirdly with looking up the ref cache… what if there are two of the same linked formats registered at the same position? I'm thinking the simpler option for might just to let things be derefed from some root buffer, but we could always revisit this stuff later.

fathom/src/core/binary.rs
wezm: I don't think it would be weird.

wezm:
> I'm thinking the simpler option for might just to let things be derefed from some root buffer, but we could always revisit this stuff later.

That might be the way to go for now, since it's the least restrictive. When we have more formats defined we'll have a better idea of whether we need to be able to deref outside the buffer. If not, then perhaps the restriction can be added then.

wezm:
Latest changes look good

brendanzab:
I think this should be good to go. I wait for #363 if you like though?

wezm:
> I think this should be good to go. I wait for #363 if you like though?

Maybe you go first. Might be easier to fix mine up since it's a bit smaller.

PR #313, opened 2022-03-24
Opentype cmap sub-table formats

wezm:
This PR describes the remaining OpenType `cmap` sub-tables as much as possible. I've tested all tables against real fonts, comparing read values to [ttf-explorer](https://github.com/RazrFalcon/ttf-explorer) except for format 13, which I couldn't find a font for.

### DIFF
--- a/formats/opentype.fathom
+++ b/formats/opentype.fathom
@@ -394,6 +394,7 @@ let cmap_subtable_format4 = fun (platform : Repr platform_id) => {
     language <- cmap_language_id platform,
     /// 2 × segCount.
     seg_count_x2 <- u16be,
+    seg_count <- succeed _ (u16_div seg_count_x2 2),
     /// Maximum power of 2 less than or equal to segCount, times 2 ((2**floor(log₂(segCount))) * 2,
     /// where “**” is an exponentiation operator)
     search_range <- u16be,
@@ -403,17 +404,16 @@ let cmap_subtable_format4 = fun (platform : Repr platform_id) => {
     /// segCount times 2, minus searchRange ((segCount * 2) - searchRange)
     range_shift <- u16be,
     /// End characterCode for each segment, last=0xFFFF.
-    end_code <- array16 seg_count_x2 u16be, // TODO: seg count / 2
+    end_code <- array16 seg_count u16be,
     /// Set to 0.
     _reserved_pad <- reserved s16be 0,
-
-    // TODO these need arithmetic
-    // /// Start character code for each segment.
-    // start_code[segCount]  <- u16be,
-    // /// Delta for all character codes in segment.
-    // id_delta[segCount]  <- int16,
-    // /// Offsets into glyphIdArray or 0
-    // id_range_offsets[segCount]  <- u16be,
+    /// Start character code for each segment.
+    start_code <- array16 seg_count u16be,
+    /// Delta for all character codes in segment.
+    id_delta <- array16 seg_count s16be,
+    /// Offsets into glyphIdArray or 0
+    id_range_offsets <- array16 seg_count u16be,
+    // TODO: Needs length limiting formats
     // /// Glyph index array (arbitrary length)
     // glyph_id_array[ ]  <- u16be,
 };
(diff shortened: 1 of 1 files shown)


formats/opentype.fathom
```diff
@@ -198,10 +207,77 @@ let language_id =
 let cmap_language_id = fun (platform : Repr platform_id) =>
     language_id;
 
+// cmap sub-table format 8 has a 32-bit language code
+let cmap_language_id32 = fun (platform : Repr platform_id) =>
+    language_id32;
```
mikeday, line 212: Why is this parameterised by platform?

wezm: Ahh I meant to comment on that. There might be a better way to do this.

It's parameterised by platform because I just copied the 16-bit version that was already there. https://github.com/yeslogic/fathom/pull/313/files#diff-fd2bf33c4274b68a3b5148dafc986eafbff4ac002bb464422b4178c92d420834R207

wezm: I think Brendan parameterised it by platform because the interpretation of the language depends on the platform but I don't think that influences parsing.

brendanzab: Yeah, it's more about how this information is interpreted (I forgot to add a TODO). See the [MS OpenType Spec](https://docs.microsoft.com/en-us/typography/opentype/spec/cmap#use-of-the-language-field-in-cmap-subtables) for more information. We could always remove it for now though.

formats/opentype.fathom
```diff
@@ -224,33 +302,228 @@ let byte_encoding_table = fun (platform : Repr platform_id) => {
...
 let character_map_subtable = fun (platform : Repr platform_id) => {
+    /// The start of the character mapping sub-table
+    table_start <- stream_pos,
     /// Format number of the subtable
     format <- u16be,
-    /// The length ot the subtable in bytes
-    length <- u16be,
-    // TODO: format-specific subtable data
     data <- match format {
         0 => byte_encoding_table platform,
```
brendanzab, line 518: Could maybe rename this to `cmap_subtable_format0` for consistency 🤔

brendanzab:
This is so cool! Looking really good!

formats/opentype.fathom
wezm: Ahh good idea. I didn't follow you lead there with the naming because the names of the other tables were a bit unwieldy and I thought it would be more readable to refer to them by their format number.

formats/opentype.fathom
```diff
@@ -224,33 +302,228 @@ let byte_encoding_table = fun (platform : Repr platform_id) => {
...
+    /// Tightly packed array of bits (8K bytes total) indicating whether the particular 16-bit
+    /// (index) value is the start of a 32-bit character code
+    is32 <- array16 8192 u8,
+    /// Number of groupings which follow
+    num_groups <- u32be,
+    /// Array of SequentialMapGroup records.
+    groups <- array32 num_groups sequential_map_group,
+};
+
+/// Format 10: Trimmed table mapping
```
brendanzab, line 419: Some of  these doc comments use inconsistent headings - not sure if it really matter at this stage though!

brendanzab (approved):

formats/opentype.fathom
brendanzab: Yeah that makes sense - I though it would be nice to have some more, uh, ‘semantic’ names for them, but I hadn't yet got up to the more goofy names. Can definitely understand going with numbered names here. I'm also willing to revisit the names I've given other tables too, if you disagree.

formats/opentype.fathom
wezm: Ahh, there's no way to extract the doc comments yet is there?

brendanzab: No :(

Was hoping one day to be able to extract a specification site. And also with the binary grep tooling you could get explanations of fields/types somehow. But yeah that is definitely a way off.

wezm: I fixed the Markdown

brendanzab: Looks good! 👍

PR #320, opened 2022-03-28
Simplify link formats and implement deref formats

brendanzab:
This PR adds a deref format for eagerly parsing data at a given position:

- `deref : Pos -> Format -> Format`

A number of primitives for adding unsigned numbers to positions are also implemented:

- `pos_add_u8 : Pos -> U8 -> Pos`
- `pos_add_u16 : Pos -> U16 -> Pos`
- `pos_add_u32 : Pos -> U32 -> Pos`
- `pos_add_u64 : Pos -> U64 -> Pos`

This means that we can simplify link formats down to a single primitive:

- `link : Pos -> Format -> Format`

## Additional details

I originally attempted to implement derefs as:

- `deref : fun (A : Type) -> Ref A -> Format`

But this proved too challenging to implement in the binary interpreter. We'd have to do something like:

- figure out if that link had been parsed or not, and if not:
    - look up the list of pending references for the format to be parsed at that location
    - check if the format had the same representation as the reference and parse it
- otherwise:
  - check that type of the data at had the same representation as the reference and return it
    
I might give it another go after I've given this some more thought.


tests/succeed/format-deref/simple.fathom
```diff
@@ -0,0 +1,5 @@
+{
+  start <- stream_pos,
+  len <- deref start u16be,
+  data <- array16 len u16be,
+}
```
mikeday, line 5: Implies that the length of the array is also its first element?

mikeday:
Now I'm a bit confused about the difference between `deref` and `link` given that they have the same types, is it just that `deref` is parsed immediately (and repeatedly if it occurs more than once?) while `link` is parsed later (and only once)?

mikeday:
Oh does `link` have a repr value of type `ref T`, which we can't dereference yet, while `deref` has a repr value of just T? :laughing:

brendanzab:
Yeah it's all a bit of a mess but you have the gist.

Links store a `(Pos, Format)` thing in some global state during parsing. At the end of parsing the root format, the pending formats with be parsed in turn, building up a map of `(Pos, (A : Type, A))`. If a pending position has already been occupied, then the representation of the format must match the type of what is stored in the map of parsed positions, otherwise it's a parse failure.

Originally I was hoping I could deref the `Ref A`s, but yeah I haven't figured it out and links just make me feel uncomfortable at any rate, at least in terms of how I've currently implemented them.

tests/succeed/format-deref/simple.fathom
brendanzab: Ahh yeah there should be a reserved thingo there

PR #321, opened 2022-03-29
Simplify link formats with position addition operations

brendanzab:
This implements a number of primitives for adding unsigned numbers to positions:

- `pos_add_u8 : Pos -> U8 -> Pos`
- `pos_add_u16 : Pos -> U16 -> Pos`
- `pos_add_u32 : Pos -> U32 -> Pos`
- `pos_add_u64 : Pos -> U64 -> Pos`

This means that we can simplify link formats down to a single primitive:

- `link : Pos -> Format -> Format`

At the moment these operations will get 'stuck' during evaluation if they overflow, which will ultimately lead to a parse error. We might need to consider a different approach in the future but I think this is fine for now.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,7 +1,0 @@
-{
-    "recommendations": [
-        "cab404.vscode-direnv",
-        "jnoortheen.nix-ide",
-        "matklad.rust-analyzer",
-    ]
-}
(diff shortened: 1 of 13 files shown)


brendanzab:
Splitting this out of #320

mikeday (approved):
Looks good!

brendanzab:
I'll merge this after #321 as it will require some snapshot updates.

Issue #318, opened 2022-03-25
Top-level items


brendanzab:
Currently a fathom file consists of a single expression. This is appealing in its simplicity, but in lieu of a fancier record-based module system (which is out of scope for Fathom), it is a pain to use in practice. The proposal is to switch to the following syntax:

```ebnf
module ::= item*

item ::= 
  | "def" ident (":" term)? "=" term ";"
```

Eventually we could extend this to support more top-level items, but this should work for now. This will unlock the following benefits:

- allow us to topologically sort definitions before elaborating them, allowing definitions to be ordered from the roots to the leaves
- allow us to more easily add imports to modules, allowing complicated formats to be broken up into smaller files
- let users read from specific top-level formats via the CLI
- make it easier to do an initial pass on glued evaluation

Issue #316, opened 2022-03-25
Remove Type-in-Type


brendanzab:
Currently Fathom implements `Type : Type`. This is a common approach for making prototype implementations of dependently typed languages, but it's far too expressive for a binary data description language. It also introduces well-known paradoxes that can lead to non-termination. Instead we should only ever allow `Type` to appear on the right-hand side of the colon. This could involve adding a `check_type` function to the elaborator:

```rust
pub fn check_type(&mut self, surface_term: &Term<'_, ByteRange>) -> core::Term<'arena> {
    ...
}
```

Similarly it might also make sense to remove the `Format : Type` rule, instead checking that `Format` is a type in the `check_type` function. This prevent us from defining binary formats that describe other binary formats, but these are probably exceedingly rare, and it's probably good to be more conservative at this point.

brendanzab:
I was looking into restricting fathom to only allow us to talk about “small types” (The Little Typer does this in its Pie language). I was running into issues with unification introducing `Type`s however.

I think that we could introduce a restricted universe hierarchy with just a universe of small types (`Type`), and a universe of large types (`Type 1`). This is what XML does (a core language for SML by Harper and Mitchell). 1ML does something similar (but has not dependent types). As in Idris there would be no syntax to introduce `Type 1` universes, but we also would not attempt to ‘solve’ universe levels to keep things simpler. Formats would be considered ‘large types’ (as they can contain types). I also think records should only be permitted to contain small types for the time being for simplicity. I’ve implemented a simple approach to checking universes in [wip-compile-stratify](https://github.com/brendanzab/language-garden/blob/main/wip-compile-stratify/Core.ml).

PR #322, opened 2022-03-29
Allow different formats at the same ref position

brendanzab:
This allows multiple linked formats to occupy the same position in the binary file.

Currently this just renders the different formats as a list corresponding to each position in the CLI. In the future I'd want to explore a different way of displaying these - perhaps something like `pos : Repr format = ...` or `pos : format = ...` (if we add format -> repr coercions) - but this would require glued evaluation of [top-level items](https://github.com/yeslogic/fathom/issues/318) to avoid swamping users with massive format descriptions.

### DIFF
--- a/fathom/src/core/binary.rs
+++ b/fathom/src/core/binary.rs
@@ -13,8 +13,12 @@ pub struct Context<'arena, 'env> {
     pending_formats: Vec<(u64, ArcValue<'arena>)>,
 }
 
-pub struct RefData<'arena> {
-    pub r#type: ArcValue<'arena>,
+pub struct ParsedRef<'arena> {
+    /// The format that this reference was parsed with
+    // Invariant: `format : Format`
+    pub format: ArcValue<'arena>,
+    /// The expression that was parsed for this reference
+    // Invariant: `expr : Repr format`
     pub expr: ArcValue<'arena>,
 }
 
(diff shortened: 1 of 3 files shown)


fathom/src/core/binary.rs
```diff
@@ -39,42 +43,26 @@ impl<'arena, 'env> Context<'arena, 'env> {
...
-                    // Record the data in the `refs` hashmap
-                    entry.insert(RefData {
-                        r#type: format_repr,
-                        expr,
-                    });
-                }
+            let parsed_refs = refs.entry(pos).or_insert(Vec::with_capacity(1));
+
+            if (parsed_refs.iter())
+                .find(|r| self.conversion_context().is_equal(&r.format, &format))
```
brendanzab, line 57: It would be nice to make this more efficient. This would be easier if we could somehow assign a unique id to formats at compile time. This is complicated due to:

- runtime data dependencies in format descriptions 
- the fact that formats can be constructed at runtime (I think #316 might resolve this)

brendanzab: Now I'm curious if Kaitai allows instances to be parametrised values 🤔

mikeday (approved):
Looks good to me!

fathom/src/core/binary.rs
brendanzab: Seems like no, Kaitai does not allow instances to have parameters, as far as I can see. Nor can they handle mutual recursion of the font tables (they seem to skip the tables where this comes up, like `htmx`?).

PR #327, opened 2022-03-30
Implement deref formats

brendanzab:
This is take two of implementing deref formats, after #320 . This alters references to now be indexed with the format description of the linked data (as opposed to the representation type):

- `Ref : Format -> Type`

The following format description allows for parsing those references eagerly:

- `deref : fun (f : Format) -> Ref f -> Format`

The number of calls to `ConversionContext::is_equal` in order to look up references in the cache is a bit of a worry, but I have some ideas about how to improve this if it ever becomes a problem, which I've noted down in comments.

Closes #320

### DIFF
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -382,7 +382,7 @@ Links formats are [represented](#format-representations) as typed
 ### Deref formats
 
 Deref formats allow [references](#references) to other parts of the stream to be
-parsed eagerly:
+included in resulting parsed output.
 
 - `deref : fun (f : Format) -> Ref f -> Format`
 
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -24,8 +24,9 @@ elaboration, and core language is forthcoming.
   - [Overlap formats](#overlap-formats)
   - [Number formats](#number-formats)
   - [Array formats](#array-formats)
-  - [Link formats](#link-formats)
   - [Stream position formats](#stream-position-formats)
+  - [Link formats](#link-formats)
+  - [Deref formats](#deref-formats)
   - [Succeed format](#succeed-format)
   - [Fail format](#fail-format)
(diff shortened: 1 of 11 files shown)


doc/reference.md
```diff
@@ -361,22 +375,25 @@ to expect at that position:
...
+| ----------------- | ------------- |
+| `link pos format` | `Ref format`  |
 
-### Stream position formats
+### Deref formats
 
-The stream position format is interpreted as the current stream position during
-parsing:
+Deref formats allow [references](#references) to other parts of the stream to be
+parsed eagerly:
```
mikeday, line 385: I'm not sure if the eagerness of the parsing is the critical part of the semantics here, like the reason a spec author might need a deref is probably not most clearly expressed as eagerness exactly.

brendanzab: Yeah I was worried when I wrote that and debated putting it in but was lazy (hah) - I'll try to fix it up tomorrow

fathom/src/core/binary.rs
```diff
@@ -208,6 +193,79 @@ impl<'arena, 'env> Context<'arena, 'env> {
...
+        reader.seek(SeekFrom::Start(pos))?;
+        // Parse the data at that location
+        let expr = self.read_format(reader, &format)?;
+        // Reset reader back to the original position
+        reader.seek(SeekFrom::Start(initial_pos))?;
+
+        // We might have parsed the current reference during the above call to
+        // `read_format` so it's important to check first, in order to avoid
+        // duplicates in the cache.
+        if let Some(parsed_ref) = self.lookup_cached_ref(pos, &format) {
```
mikeday, line 253: I think this should be an assertion failure as it should not ever be possible for this to happen as it would imply infinite regress.

PR #330, opened 2022-04-01
Introduce RigidEnvBuilder for defining default env

wezm:
This is a refactoring in order to make defining the boolean primitives possible/easier as discussed in the chat. The naming is open to bike shedding/improvement, feel free to make suggestions.


brendanzab (approved):
Looks good! Eventually I want to move this to the core and give it a bit of a rethink, but I think this is good for cleaning things up in the mean time. 👍

PR #332, opened 2022-04-04
Clean up grammar with sequence macro

brendanzab:


wezm (approved):
Looks like a good clean up

PR #333, opened 2022-04-04
Restrict where pattern annotations can appear

brendanzab:
This removes annotated patterns from the pattern syntax, which avoids some some strange, potentially confusing edge-cases.


wezm (approved):
Looks good 👍

PR #334, opened 2022-04-05
Booleans and comparison ops

wezm:
This PR adds:

- Booleans
- Boolean operators
- Comparison operators


mikeday:
Can we and a u16 with another and check if the result is non-zero now? :nerd_face:

wezm:
> Can we and a u16 with another and check if the result is non-zero now? 🤓

Yes I think so: 

```
$ echo 'u16_neq (u16_add 1 2) 0' | cr norm
true : Bool
```

brendanzab (approved):
Wonderful 👍

wezm:
I noticed something else that I should include in this PR: an update to the reference. I'll add that at some point today.

wezm:
Now includes update to the reference.

doc/reference.md
```diff
@@ -567,6 +569,21 @@ some-data.data : Array3 some-data.len S32
 //                                      in type of the projected term
 ```
 
+## Booleans
+
+The boolean type is: `Bool` with values `true` and `false`.
```
brendanzab, line 574: I must admit I never really know when to use the term ‘value’ now, but I think this is ok given that these terms evaluate to these values? Might be something to ask type theory twitter about heh 🤔

brendanzab (approved):
Looks good!

PR #336, opened 2022-04-06
Improve implementation docs and tweak some variant names

brendanzab:


fathom/src/core.rs
```diff
@@ -6,24 +6,17 @@ use crate::StringId;
...
 //
 // See also: https://en.wikipedia.org/wiki/Abstract_and_concrete
 #[derive(Debug, Copy, Clone)]
 pub enum EntryInfo {
-    Concrete,
-    Abstract,
+    /// The entry was bound as a definition in the environment.
+    Def,
+    /// The entry was bound as a parameter in the environment
+    Param,
```
wezm, line 19: Any reason not to use the full words `Definition` and `Parameter`?

wezm (approved):
The explanations are good addition 👍

fathom/src/core.rs
brendanzab: Was mirroring the method names in `RigidEnv`, but I'm all for writing things in full!

PR #337, opened 2022-04-07
Wire up error recovery in the parser.

brendanzab:
This allows for further parsing to take place after a syntax error is encountered. For more information check out the [LALRPOP book's chapter on error recovery](https://lalrpop.github.io/lalrpop/tutorial/008_error_recovery.html).


wezm (approved):
Awesome! This should be a good usability improvement.

Issue #341, opened 2022-04-20
How to handle names in patterns


wezm:
Prior to the introduction of booleans in patterns, a name in a pattern always created a new binding that acts a catch-all. The initial implementation of booleans defines `true` and `false` in the rigid environment. For these to work in patterns we need to check if a name is defined in the environment and retrieve its value.

There is currently a hack in place to only do this for the names "true" and "false". Without this hack the value of any binding in the environment can be used in a pattern, which can lead to confusion as to when a new binding is created versus matching against the value of an existing one.

Some possible resolutions to this that @brendanzab proposed are:

- special casing true and false (and preventing matching on user-specified definitions) ← this is what we do now
- making variable patterns explicitly different to constant patterns (eg. $x or using capitalisation like in Mercury, Haskell, or OCaml) - this would have a big impact on what Fathom looks like
- making constant patterns different to variable patterns (eg. ^x like in Elixir or perhaps =x)
- just not pursuing pattern matching on constants for now, instead adding if as surface syntax for booleans (which we'll want anyway)

This issue aims to capture this for later consideration.

Kmeakin:
Resolved by https://github.com/yeslogic/fathom/pull/343?

brendanzab:
It’s… kind of a solution. I guess if we don’t want to be super exotic then “doing what Rust does” might work? 🤔

Kmeakin:
The "doing what Rust does" solution seems like a practical low-tech solution: in 99% of situations, you can distinguish between variable bindings and unit enum/struct patterns by capitalization. Restrictions on which identifiers can be used in which position (a la Haskell) could make codegen by macros/tools more difficult.

PR #342, opened 2022-04-20
Bool patterns

### LINKED ISSUE #340: Check for duplicate patterns in match
When checking constant patterns we don't report unreachable duplicates:

https://github.com/yeslogic/fathom/blob/6b9d1af424272b719f1cd64de8a17051860d952a/fathom/src/surface/elaboration.rs#L1672-L1678

In the forthcoming boolean pattern support duplicate patterns can result in a panic if there is exactly two branches matching on the same value. This makes it appear as an exhaustive match but later in [apply_const](https://github.com/yeslogic/fathom/blob/6b9d1af424272b719f1cd64de8a17051860d952a/fathom/src/core/semantics.rs#L636) the absence of a default pattern leads to a panic.

Noting this in an issue for follow-up after bool matches is done.

wezm:
This PR adds initial support for boolean patterns. As a demonstration of their use I extended `opentype.fathom` to implement some of the version and magic number checks.

There are several things that have been deemed out of scope for this PR that will be investigated/addressed in follow up work:

- The check for sfnt version is commented out as it has a big impact on run time, although it does work.
- There are two FIXMEs attributed to #340
- When encountering a name in a pattern there is a special case to handle `true` and `false` to prevent all binding from resolving to their value. #341 captures the alternatives to this for later resolution.

**Other notes:**

I'm open to better name suggestions for `SplitConstBranches` and its variants.

Fixes #340


mikeday:
I'm curious why the sfnt version check has an impact when the other tests do not?

wezm:
> I'm curious why the sfnt version check has an impact when the other tests do not?

Looks like I've found what to look at next :)

wezm:
You know what I think it was the test harness that was going slow. Now that I've updated the snapshots for the other differences it's not slow. Perhaps generating the diffs was slow for some reason.

fathom/src/core/semantics.rs
```diff
@@ -181,6 +181,13 @@ impl<'arena> Split<'arena> {
     }
 }
 
+#[derive(Clone, Debug)]
+pub enum SplitConstBranches<'arena> {
+    Const(Const, ArcValue<'arena>, Split<'arena>),
+    Default(Closure<'arena>),
+    None,
+}
```
brendanzab, line 189: Maybe it would help reduce confusion to rename `Split` to `ConstBranches`? I was stuck in naming this originally, and it shows in some of the identifiers I picked. 

It's confusing because in`split_telescope` and `split_const_branches` I was trying to follow in an attempt to follow Rust's name for ‘uncons’, [`[T]::split_first`](https://doc.rust-lang.org/std/primitive.slice.html#method.split_first). But then this collides with ‘split’ in the sense of a ‘case-split’.

Also wondering if we could make this type of `split_const_branches` follow [`[T]::split_first`](https://doc.rust-lang.org/std/primitive.slice.html#method.split_first) more closely with something like:

```rs
Option<((Const, ArcValue<'arena>), NextConstBranches<'arena>)>
```

using the following type:

```rs
#[derive(Clone, Debug)]
pub enum NextConstBranches<'arena> {
    Branches(ConstBranches<'arena>),
    Default(Closure<'arena>),
}
```

I dunno if that's better though - just an idea!

fathom/src/core/semantics.rs
```diff
@@ -650,7 +662,13 @@ impl<'arena, 'env> ElimContext<'arena, 'env> {
                 // Otherwise call default with `head_expr`
                 let mut rigid_exprs = split.rigid_exprs.clone();
                 rigid_exprs.push(head_expr);
-                EvalContext::new(&mut rigid_exprs, self.flexible_exprs).eval(split.default_expr)
+                match split.default_expr {
+                    Some(default_expr) => {
+                        EvalContext::new(&mut rigid_exprs, self.flexible_exprs).eval(default_expr)
+                    }
+                    // FIXME: https://github.com/yeslogic/fathom/issues/340
+                    None => panic_any(Error::MissingConstDefault),
```
brendanzab, line 670: I think it's fine to add the FIXME here - but when #340 is fixed this panic will probably remain - this should be caught by the elaborator when compiling the pattern matches.

fathom/src/core/semantics.rs
```diff
@@ -787,18 +805,21 @@ impl<'in_arena, 'out_arena, 'env> QuoteContext<'in_arena, 'out_arena, 'env> {
...
+                                }
+                                SplitConstBranches::None => break None,
                             }
                         };
 
                         Term::ConstElim(
                             self.scope.to_scope(head_expr),
                             branches.into(),
-                            self.scope.to_scope(default_expr),
+                            default_expr.map(|expr| &*self.scope.to_scope(expr)),
```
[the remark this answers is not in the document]
wezm: This and the one above were to appease compiler. Without it, it results in this error:

```
error[E0308]: mismatched types
   --> fathom/src/core/semantics.rs:822:29
    |
822 | ...                   default_expr.map(|expr| self.scope.to_scope(expr)),
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ types differ in mutability
    |
    = note: expected enum `Option<&core::Term<'_>>`
               found enum `Option<&mut core::Term<'_>>`
```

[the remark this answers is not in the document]
brendanzab: Ohhhhhh got it! Yeah I think I've done `as &_` in the past to show that the reference is being cast to a shared ref, as opposed to dereffing and reffing. Not sure what is better though.

brendanzab:
Looking good! Thanks for the various comment fixes by the way! 😄

fathom/src/core/semantics.rs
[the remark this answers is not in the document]
wezm: `as &_` definitely seems to express the intent better so I'll change to that.

fathom/src/surface/elaboration.rs
```diff
@@ -830,7 +834,20 @@ impl<'interner, 'arena> Context<'interner, 'arena> {
...
+                    .filter(|&name| name == "true" || name == "false")
+                    .and_then(|_| self.get_name(*name))
+                    .and_then(|(local_var, _type)| self.get_name_def(local_var))
+                    .and_then(|value| match **value {
+                        Value::Const(r#const) => Some(CheckedPattern::Const(*range, r#const)),
+                        _ => None,
+                    })
+                    .unwrap_or_else(|| CheckedPattern::Name(*range, *name));
+
+                (checked_pattern, expected_type.clone())
```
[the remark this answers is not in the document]
wezm: lol yes, that's much nicer.

fathom/src/core/semantics.rs
wezm: Ahh interesting. That explanation does connect some dots for me. Perhaps `Split` could also be `CaseSplit` but `ConstBranches` is does seem clearer.

> Also wondering if we could make this type of `split_const_branches` follow [`[T]::split_first`](https://doc.rust-lang.org/std/primitive.slice.html#method.split_first) more closely with something like:
> 
> ```rust
> Option<((Const, ArcValue<'arena>), NextConstBranches<'arena>)>
> ```

This also makes more sense now. I didn't understand in the original code why the tuples were nested versus using a 3-tuple. Perhaps it's worth giving `(Const, ArcValue<'arena>)` a name, even if only an alias like `Branch` or `ConstBranch` to clean up the nesting a bit?

brendanzab: Yeah good idea about giving it a name!

wezm: I had a go at refactoring this to use

```rust
Option<((Const, ArcValue<'arena>), NextConstBranches<'arena>)>
```

but it got a bit messy. Perhaps I misunderstood something but I decided to mostly leave it alone. I did change 

```rust
SplitConstBranches::Const(Const, ArcValue<'arena>, Split<'arena>)
```

to

```rust
SplitConstBranches::Branch(Branch<'arena>, Split<'arena>)
```

though, where `Branch<'arena>` is `type Branch<'arena> = (Const, ArcValue<'arena>);`

brendanzab: Yeah that's fair enough - I might have a go later if that's ok.

Eventually might be good to rename `Split` to `Branches` but it's probably good to take one thing at a time.

tests/fail/elaboration/non-exhaustive-patterns/match-duplicate.snap
```diff
@@ -0,0 +1,19 @@
...
+  ┌─ tests/fail/elaboration/non-exhaustive-patterns/match-duplicate.fathom:5:3
+  │
+5 │   true => 2
+  │   ^^^^
+
+error: non-exhaustive patterns in match expression
+  ┌─ tests/fail/elaboration/non-exhaustive-patterns/match-duplicate.fathom:3:7
+  │  
+3 │ ╭ match true {
+  │         ^^^^ patterns not covered
```
brendanzab, line 13: This is a bug in codespan. I'm pretty sure it was fixed but we might not have deployed it to crates.io

brendanzab (approved):
Looks good!

fathom/src/core/semantics.rs
wezm: > Yeah that's fair enough - I might have a go later if that's ok.

By all means go ahead 👍

tests/fail/elaboration/non-exhaustive-patterns/match-duplicate.snap
wezm: What's buggy about it?

brendanzab: It's missing the vertical line in the gutter

brendanzab:
```
  ┌─ tests/fail/elaboration/non-exhaustive-patterns/match-duplicate.fathom:3:7
  │  
3 │ ╭ match true {
  │ │       ^^^^ patterns not covered
  ```

PR #344, opened 2022-04-22
Add formatting hints to numbers

### LINKED ISSUE #328: Add hints to `Const`s to improve the distillation of integer constants
Currently the distiller has know way of knowing how to render constants, instead just defaulting to rendering as decimal numbers. We could fix this by altering `Const` to be:

```rs
enum IntStyle {
    Binary,
    Decimal,
    Hexadecimal,
    Ascii,
}

pub enum Const {
    U8(u8, IntStyle),
    U16(u16, IntStyle),
    U32(u32, IntStyle),
    U64(u64, IntStyle),
    S8(i8, IntStyle),
    S16(i16, IntStyle),
    S32(i32, IntStyle),
    S64(i64, IntStyle),
    // TODO: use logical equality for floating point numbers
    F32(f32),
    F64(f64),
    Pos(u64),
    Ref(u64),
}
```

This will allow us to convert back to an appropriate literal (either numeric or string) during distillation.

One thing we'll need to consider is to ignore the style hint when comparing constants for equality. We could do this by a custom overload `PartialEq`, or, by implementing our own method (eg. `Const::logical_eq`), which will involve updates to `surface::elaboration::unification` and `core::semantics`:

- https://github.com/yeslogic/fathom/blob/76414023a3e7298be1b1386c23a37e77b80135de/fathom/src/core/semantics.rs#L948
- https://github.com/yeslogic/fathom/blob/76414023a3e7298be1b1386c23a37e77b80135de/fathom/src/surface/elaboration/unification.rs#L280

Another complication is how to handle the style hints in the primitive/builtin ops - we'll need to ignore or merge them somehow to figure out what to use in the returned constant.

wezm:
Fixes #328

Opening this PR for feedback, before proceeding further. The PR adds int hints so that number literals and number patterns retain their original format. For example:

```
$ echo 'let x : U32 = 1234; let _x : U16 = match x { 0x1234 => 1, 0b1111 => 2, "head" => 4,  _ => 3 }; _x' | mold -run cargo r -- elab
let x : U32 = 1234;
let _x : U16 = match x { 0b1111 => 2, 0x1234 => 1, "head" => 4, _ => 3 };
_x : U16
```

**Possible further work**

- ~~I think the extra `IntStyle` field in `Const` messes up `PartialEq`. Looks like I'll need to manually implement it and have it ignore that field.~~ — I've now done this
  - Oh you totally covered this in the issue I linked "One thing we'll need to consider is to ignore the style hint when comparing constants for equality. We could do this by a custom overload PartialEq, or, by implementing our own method (eg. Const::logical_eq), which will involve updates to surface::elaboration::unification and core::semantics"
- A way to add hints to binary data that is read, for example indicating that `table_id` in `table_record` in `opentype.fathom` should use the ASCII format.
- Pehaps find a better home for the traits etc. that were added to `core`.
- ~~Propagate the style from the inputs to ops~~ — I had a go at this
- ~~Add some more tests~~
- Allow hex/bin signed ints and extend this to them
- The ascii format uses big endian... should that be adjustable some how?


fathom/src/core.rs
```diff
@@ -422,6 +431,88 @@ pub enum Const {
...
+
+pub trait Styled<const N: usize>:
+    std::fmt::Display + Copy + std::fmt::LowerHex + std::fmt::Binary + ToBeBytes<N>
+{
+}
+
+impl IntStyle {
+    pub fn format<T: Styled<N>, const N: usize>(&self, number: T) -> String {
+        match self {
+            IntStyle::Binary => format! {"0b{:b}", number},
```
brendanzab, line 483: Why do we use braces for some of these `format!` invocations? Is that for some sort of escaping reason?

fathom/src/core.rs
```diff
@@ -422,6 +431,88 @@ pub enum Const {
...
+        }
+
+        impl Styled<{std::mem::size_of::<$ty>()}> for &$ty {}
+        )*
+    };
+}
+
+impl_styled_int!(u8, u16, u32, u64);
+
+pub trait Styled<const N: usize>:
```
brendanzab, line 475: Dunno if better name could be `IntStyled` or `UIntStyled`?

fathom/src/core.rs
```diff
@@ -403,14 +403,23 @@ def_prims! {
     PosAddU64 => "pos_add_u64",
 }
 
-/// Constants
+/// Formatting style for integers
 #[derive(Debug, Copy, Clone, PartialEq, PartialOrd)]
+pub enum IntStyle {
```
brendanzab, line 408: Not sure if `UIntStyle` is a better name, given it is only used for unsigned integers. Not a big issue though - we can always defer this until we extend support to Signed integers as well.

brendanzab (commented):
Looking good so far! I really like the merging used in the primops - it's nice to try to preserve the styling where possible.

Good point about the endianness of the ASCII literals. I wasn't really sure about them when I first implemented them, but yeah, they might require some more thought. 🤔

fathom/src/core.rs
```diff
@@ -422,6 +431,88 @@ pub enum Const {
...
+        }
+
+        match (self, other) {
+            // If one is decimal then return the other
+            (Decimal, _) => other,
+            (_, Decimal) => self,
+            // otherwise favour ourself
+            (Hexadecimal, _) | (Binary, _) | (Ascii, _) => self,
+        }
+    }
```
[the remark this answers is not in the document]
wezm: > seems overly biased towards self

Absolutely. This was my lack of a better idea. Your suggestion is much better, thanks.

fathom/src/core/semantics.rs
```diff
@@ -384,67 +390,67 @@ fn prim_step(prim: Prim) -> Option<PrimStep> {
...
-        Prim::U8Sub => const_step!([x: U8, y: U8] => Const::U8(u8::checked_sub(*x, *y)?)),
-        Prim::U8Mul => const_step!([x: U8, y: U8] => Const::U8(u8::checked_mul(*x, *y)?)),
-        Prim::U8Div => const_step!([x: U8, y: U8] => Const::U8(u8::checked_div(*x, *y)?)),
-        Prim::U8Not => const_step!([x: U8] => Const::U8(u8::not(*x))),
-        Prim::U8Shl => const_step!([x: U8, y: U8] => Const::U8(u8::checked_shl(*x, u32::from(*y))?)),
-        Prim::U8Shr => const_step!([x: U8, y: U8] => Const::U8(u8::checked_shr(*x, u32::from(*y))?)),
-        Prim::U8And => const_step!([x: U8, y: U8] => Const::U8(u8::bitand(*x, *y))),
-        Prim::U8Or => const_step!([x: U8, y: U8] => Const::U8(u8::bitor(*x, *y))),
-        Prim::U8Xor => const_step!([x: U8, y: U8] => Const::U8(u8::bitxor(*x, *y))),
+        Prim::U8Add => const_step!([x, xst: U8, y, yst: U8] => Const::U8(u8::checked_add(*x, *y)?, xst.merge(*yst))),
```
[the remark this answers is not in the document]
wezm: I'm not fussed either way. I think I actually started with `IntStyle::merge` but happy to change it back.

fathom/src/core.rs
wezm: I suspect I just mistyped `(` and being a macro it still works. It wasn't intentional, will fix.

fathom/src/core.rs
```diff
@@ -403,14 +403,23 @@ def_prims! {
...
 }
 
-/// Constants
+/// Formatting style for integers
 #[derive(Debug, Copy, Clone, PartialEq, PartialOrd)]
+pub enum IntStyle {
+    Binary,
+    Decimal,
+    Hexadecimal,
+    Ascii,
```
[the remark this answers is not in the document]
wezm: A little research suggests that the most common case is for these to be big-endian (E.g. [OSType](https://en.wikipedia.org/wiki/FourCC), PNG, AIFF, RIFF, AVI, WAV, DirectX, OpenType). Seems reasonable to just call it `Ascii` for now.

wezm:
Ok I addressed the comments. I originally opened this PR with some parts remaining but many have been addressed. So it's probably suitable for merging now, if you agree.

brendanzab (approved):
Looks good!

PR #348, opened 2022-04-27
Simplify RigidEnvBuilder

brendanzab:
This simplifies the `RigidEnvBuilder` by constructing terms then evaluating them, rather than constructing values directly.

### DIFF
--- a/fathom/src/surface/elaboration.rs
+++ b/fathom/src/surface/elaboration.rs
@@ -94,69 +94,69 @@ impl<'arena> RigidEnv<'arena> {
         const S64_TYPE: core::Term<'_> = core::Term::Prim(Prim::S64Type);
         const POS_TYPE: core::Term<'_> = core::Term::Prim(Prim::PosType);
 
-        let mut def = RigidEnvBuilder::new(interner, scope);
+        let mut env = RigidEnvBuilder::new(interner, scope);
 
-        def.define_prim(VoidType, &UNIVERSE);
+        env.define_prim(VoidType, &UNIVERSE);
 
-        def.define_prim(BoolType, &UNIVERSE);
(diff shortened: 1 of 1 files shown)


fathom/src/surface/elaboration.rs
```diff
@@ -80,207 +80,224 @@ impl<'arena> RigidEnv<'arena> {
...
+        const U16_TYPE: core::Term<'_> = core::Term::Prim(Prim::U16Type);
+        const U32_TYPE: core::Term<'_> = core::Term::Prim(Prim::U32Type);
+        const U64_TYPE: core::Term<'_> = core::Term::Prim(Prim::U64Type);
+        const S8_TYPE: core::Term<'_> = core::Term::Prim(Prim::S8Type);
+        const S16_TYPE: core::Term<'_> = core::Term::Prim(Prim::S16Type);
+        const S32_TYPE: core::Term<'_> = core::Term::Prim(Prim::S32Type);
+        const S64_TYPE: core::Term<'_> = core::Term::Prim(Prim::S64Type);
+        const POS_TYPE: core::Term<'_> = core::Term::Prim(Prim::PosType);
+
         let mut def = RigidEnvBuilder::new(interner, scope);
```
wezm, line 97: Minor: Might be worth renaming the builder now? I originally named it this so that a line like `def.format(FormatU64Le);` read like `define format FormatU64Le` but now define is in the `define_prim` method call. Maybe call it `env` now?

wezm (approved):
Looks like a nice simplification 👍

fathom/src/surface/elaboration.rs
brendanzab: Oh good idea

PR #350, opened 2022-04-27
Add part of glyf and loca to OpenType format

wezm:
I made a start on `glyf` and `loca`. To test what I did (all the `aots` fonts are CFF) I imported some simple test fonts from the WOFF/WOFF2 test suites.


wezm:
To properly parse `glyf` it should be driven by `loca`. We need a way to iterate over pairs of `loca` entries, which would be used to specify the starting and ending offsets of the glyph within the `glyf` table.

formats/opentype.fathom
```diff
@@ -1002,6 +1004,120 @@ let naming = {
...
+        true => match (args_are_signed flags) {
+            true => s16be,
+            false => u16be,
+        },
+        // Otherwise they are 8-bit
+        false => match (args_are_signed flags) {
+            true => s8,
+            false => u8,
+        },
+    };
```
brendanzab, line 1084: This is pretty cool! :)

brendanzab (approved):
Looks good!

PR #351, opened 2022-04-28
Define remaining required OpenType tables

wezm:
This adds:

- `OS/2`
- `post`
- Enable the name records in the `name` table.

`post` version 2 (embedded Pascal string glyph names) does not currently parse the names. I think this would be possible if there was a way to get the maximum (numeric) value from an array. That maximum value is the number of glyph names to read, which would allow something like

```
let pascal_string = {
    length <- u8,
    string <- array8 length u8
};

string_data <- array16 num_strings pascal_string,
```

Also is there a better way to handle the layering of versions in `OS/2`?


mikeday:
It might be time to make an issue for "sequences of unspecified length", with the glyph names example and also the `glyf` flags example and others as we encounter them.

formats/opentype.fathom
```diff
@@ -1121,7 +1118,177 @@ let glyph_data = fun (num_glyphs : U16) => {
...
+        /// Version 2.5 (deprecated), offset from standard Macintosh index
+        0x00025000 => {
+            /// Number of glyphs
+            num_glyphs <- u16be,
+            /// Difference between graphic index and standard order of glyph
+            offset <- array16 num_glyphs s8,
+        },
+        /// Version 3, no glyph names stored in font
+        0x00030000 => {},
+        _ => fail,
```
brendanzab, line 1284: Wondering if we should fail here, or just leave it open to future expansion… (not that I'd really expect anything to happen in the future with postscript).

wezm: I suppose we could return no names in the case of an unknown version. `OS/2` has a similar issue, although in that case we can probably parse the newest version we know about.

brendanzab:
> It might be time to make an issue for "sequences of unspecified length", with the glyph names example and also the `glyf` flags example and others as we encounter them.

I did have #310 open which might do part of what we need, but I wasn't entirely happy with how I implemented the binary parser portion of it. Haven't got back to it yet.

brendanzab (approved):

wezm:
> It might be time to make an issue for "sequences of unspecified length", with the glyph names example and also the `glyf` flags example and others as we encounter them.

I opened  #352 for this.

formats/opentype.fathom
```diff
@@ -1314,8 +1489,8 @@ let table_directory = fun (file_start : Pos) => {
                     maxp.num_glyphs,
             },
             name <- required_table "name" name_table,
-            os2 <- required_table "OS/2" unknown_table,
-            post <- required_table "post" unknown_table,
+            os2 <- required_table_with_len "OS/2" os2_table,
```
wezm, line 1492: @brendanzab I rebased on `main`. I added this new function because `OS/2` needs access to the table length. Not sure if this is the best way or not.

brendanzab: Oh cool, yeah I think this is ok for now! 👍

Issue #258, opened 2020-11-05
Refinement types


brendanzab:
Refinement types are an important part of what we hope to add to Fathom. These should allow us to avoid the overflow and out-of-bounds access errors that are a common source of bugs and security vulnerabilities in hand-written binary parsers. They allow you to refine types with predicates that can be automatically discharged by an external solver like [Z3](https://github.com/Z3Prover/z3) or [CVC4](cvc4.github.io).

This is an example of an integer refined by a predicate:

```
{ x : Int | 0 <= x && x < 256 }
```

## Resources

Examples of languages with support for this refinement types are:

- [Liquid Haskell](https://github.com/FStarLang/FStar/)
- [F*](https://github.com/FStarLang/FStar/)
- [Wuffs](https://github.com/google/wuffs/)
- [Dafny](https://github.com/dafny-lang/dafny)
- [ATS](https://github.com/githwxi/ATS-Postiats)
- [Why3]()

There is a tutorial on implementing such a system:

- [Refinement Types: A Tutorial](https://arxiv.org/abs/2010.07763)
- [ranjitjhala/sprite-lang](https://github.com/ranjitjhala/sprite-lang)

Some relevant Rust crates:

- [rsmt2](https://lib.rs/crates/rsmt2)
- [z3](https://lib.rs/crates/z3)
- [rlfsc](https://lib.rs/crates/rlfsc)

## Challenges

We'll need some way to statically enforcing the subset of terms that can be sent to the solver. I'm thinking we could have another universe for this? Perhaps `Smt` or `Pred`?  I haven't thought to carefully about the details though. F* uses effects for this but I'm hoping we can come up with a simpler solution.

It would also be nice to get a list of the lemmas we need to discharge at the end of elaboration. That way we could solve them in a way that is independent of a specific tool. Perhaps this could be returned using some standard format, like [SMT-LIB 2](http://smtlib.cs.uiowa.edu)?

brendanzab:
This looks potentially related [Compiling Higher-Order Specifications to SMT Solvers:
How to Deal with Rejection Constructively](https://bentnib.org/posts/2023-01-17-higher-order-specs-to-smt.html). Seems to be about producing good errors when a predicate cannot be translated to an SMT solver?

PR #355, opened 2022-05-03
Coerce format descriptions into representation types

brendanzab:
This PR adds implicit coercions from format descriptions into their representation types during elaboration, removing the need to explicitly call `Repr` in most cases. This would _almost_ bring us full circle back to when I was trying to use subtyping in Fathom, but now the subtyping would be kept out of the core language, greatly reducing the confusing complications that resulted from conflating types and format descriptions. 

Introducing this style of coercive subtyping during elaboration is also not entirely unheard of either: it is a common addition for elaborators to type theories with Tarski-style universes, which seem to have [a connection](https://github.com/yeslogic/fathom/blob/main/doc/background.md#format-descriptions-as-universes) to our approach to format descriptions.

## Drawbacks

By inserting calls to `Repr` implicitly, people learning Fathom could end up finding it harder to understand the difference between format descriptions and host representation types. `Repr` might also show up in error messages, and this could also result in confusion if people using Fathom aren't used to being explicit about converting between them. We could try to remove the explicit `Repr` calls during distillation, but would most likely be an imperfect process and could further exacerbate confusion when they do eventually appear.

Introducing subtyping could also make unification more difficult. I don't think this is too much of an issue for this limited use case, but I still don't fully understand the implications or potential interactions. We also probably want to add more subtyping in the future, for example as a way make [refinement types](https://github.com/yeslogic/fathom/issues/258) usable (coercing from refined to non-refined terms), so we’ll probably have to deal with this anyway down the line.

## Resources

When implementing this, I was inspired a bit by [elaboration-zoo/experiments/univ-lifts](https://github.com/AndrasKovacs/elaboration-zoo/tree/401f185c9992ec6ff604989e0508f0d278eb07ab/experiments/univ-lifts)'s [`coe`](https://github.com/AndrasKovacs/elaboration-zoo/blob/401f185c9992ec6ff604989e0508f0d278eb07ab/experiments/univ-lifts/Main.hs#L262-L320) function.

### DIFF
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -22,6 +22,7 @@ elaboration, and core language is forthcoming.
 - [Formats](#formats)
   - [Format types](#format-types)
   - [Format representations](#format-representations)
+  - [Format coercions](#format-coercions)
   - [Record formats](#record-formats)
   - [Conditional formats](#conditional-formats)
   - [Overlap formats](#overlap-formats)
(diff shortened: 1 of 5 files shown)


mikeday:
That looks convenient!

brendanzab:
Thinking I might merge this - we can always remove it if it’s too confusing.

PR #357, opened 2022-05-04
Display expected and actual types when mismatched

wezm:
This is an attempt at improving the error messages when there is a type mismatch. Now the types involved are printed as well as the "type mismatch" error.

A note on the changes to the `driver`:

```rust
let err_scope = scoped_arena::Scope::new();
let mut context = elaboration::Context::new(&self.interner, &self.core_scope, &err_scope);
```

Ideally/my intention was that `&self.surface_scope` would be passed into `Context::new`, however that prevents the `.reset()` call on the scope from working a few lines down. So I pass in a separate scope.

I also added two variants to `unification::Error`, which are only created temporarily and are used to signal to higher level code that has access to the `Values` to create a `Mismatch` error with those values. I'm not sure I love this as a solution so comments welcome.


fathom/src/surface/elaboration/unification.rs
```diff
@@ -301,17 +311,19 @@ impl<'arena, 'env> Context<'arena, 'env> {
         &mut self,
         spine0: &[Elim<'arena>],
         spine1: &[Elim<'arena>],
-    ) -> Result<(), Error> {
+    ) -> Result<(), Error<'arena>> {
         if spine0.len() != spine1.len() {
-            return Err(Error::Mismatch);
+            return Err(Error::SpineMismatch);
         }
```
brendanzab, line 317: This could be moved after the for loop if we wanted to pick up on errors in the spine more eagerly.

fathom/src/surface/elaboration/unification.rs
```diff
@@ -212,15 +216,20 @@ impl<'arena, 'env> Context<'arena, 'env> {
...
                 self.unify_spines(spine0, spine1)
+                    .map_err(map_spine_mismatch(value0, value1))
             }
             (
                 Value::Stuck(Head::RigidVar(var0), spine0),
                 Value::Stuck(Head::RigidVar(var1), spine1),
-            ) if var0 == var1 => self.unify_spines(spine0, spine1),
+            ) if var0 == var1 => self
+                .unify_spines(spine0, spine1)
+                .map_err(map_spine_mismatch(value0, value1)),
```
brendanzab, line 226: What was the reasoning behind mapping to a `Mismatch` here?

fathom/src/surface/elaboration/unification.rs
```diff
@@ -301,17 +311,19 @@ impl<'arena, 'env> Context<'arena, 'env> {
...
         for (elim0, elim1) in Iterator::zip(spine0.iter(), spine1.iter()) {
             match (elim0, elim1) {
                 (Elim::Fun(input_expr0), Elim::Fun(input_expr1)) => {
                     self.unify(input_expr0, input_expr1)?;
                 }
                 (Elim::Record(label0), Elim::Record(label1)) if label0 == label1 => {}
-                (_, _) => return Err(Error::Mismatch),
+                (_, _) => {
+                    return Err(Error::SpineMismatch);
+                }
```
brendanzab, line 326: This could be an `Error::ElimMismatch(elim0, elim1)`, perhaps if we were planning on preserving more information in the unification errors.

fathom/src/surface/elaboration/unification.rs
wezm: That was the original behaviour in `unify_spines`:

```rust
        if spine0.len() != spine1.len() {
            return Err(Error::Mismatch);
```

fathom/src/surface/elaboration/reporting.rs
```diff
@@ -307,13 +306,18 @@ impl Message {
...
+                error,
+            } => {
                 use unification::{Error, RenameError, SpineError};
 
                 // TODO: Make these errors more user-friendly
                 match error {
-                    Error::Mismatch => Diagnostic::error()
-                        .with_message("type mismatch")
+                    Error::Mismatch(_, _) => Diagnostic::error()
+                        .with_message(format!("type mismatch, expected: {}, got {}", lhs, rhs))
```
brendanzab, line 320: Is there any reason to store the values here, seeing as they are not used at the mometn? Was it for improved error reporting in the future? (which I can get behind, just curious)

fathom/src/surface/elaboration/reporting.rs
```diff
@@ -87,10 +87,9 @@ pub enum Message {
     /// Unification errors.
     FailedToUnify {
         range: ByteRange,
-        // TODO: add lhs and rhs values
-        // lhs: Doc<_>,
-        // rhs: Doc<_>,
-        error: unification::Error,
+        lhs: String,
+        rhs: String,
+        error: unification::Error<'core>,
```
brendanzab, line 92: I think I had envisioned translating to a surface level unification error, seeing as value can't really be nicely rendered without the names that were bound at the point where the values were unified. But yeah, that might be pointless busywork, not sure.

fathom/src/surface/elaboration/reporting.rs
wezm: My understanding/reasoning is it's due to the core/surface separation. `Error::Mismatch` is created in the core language and it has to carry the values up to where the `Message` is created from it and the values pretty printed:

```
  match self.unification_context().unify(type0, type1) {
            Ok(()) => expr,
            Err(error) => {
                let lhs = self.pretty_print_value(type0);
                let rhs = self.pretty_print_value(type1);
                self.push_message(Message::FailedToUnify {
                    range,
                    lhs,
                    rhs,
                    error,
                });
                core::Term::Prim(Prim::ReportedError)
            }
        }
```

Ideally I think we'd drop the `error` field from `FailedToUnify` since it's no longer needed in this case after the values are pretty printed. However, the code at this point is matching on the error for other reasons unification can fail. Perhaps a new variant needs to be added to `Message` to handle this case separately?

fathom/src/surface/elaboration/unification.rs
wezm: How would we go about using `elim0, elim1` to provide more information? Part of the reason I was exploring tests for test cases was to be able to see the errors produced in these cases, but as you know that proved difficult.

wezm:
Hmm so it seems that throughout this piece of work I got myself twisted up in knots and a bunch of it remained in the code unnecessarily. Thanks for pointing out the weird things. I've pushed an update that simplifies things.

fathom/src/surface/elaboration/reporting.rs
```diff
@@ -307,13 +306,18 @@ impl Message {
...
+                error,
+            } => {
                 use unification::{Error, RenameError, SpineError};
 
                 // TODO: Make these errors more user-friendly
                 match error {
                     Error::Mismatch => Diagnostic::error()
-                        .with_message("type mismatch")
+                        .with_message(format!("type mismatch, expected: {}, got {}", lhs, rhs))
                         .with_labels(vec![Label::primary(file_id, *range)]),
```
brendanzab, line 321: Wondering what your thoughts would be on tweaking this to be something like Rust's output:

```
error[E0308]: mismatched types
   --> fathom/src/core/semantics.rs:303:27
    |
303 |                 self.eval("t_type"),
    |                           ^^^^^^^^ expected enum `core::Term`, found `str`
    |
    = note: expected reference `&core::Term<'_>`
               found reference `&'static str`
```

This would look like:

```rs
Error::Mismatch => Diagnostic::error()
    .with_message("mismatched types")
    .with_labels(vec![
        Label::primary(file_id, *range)
            .with_message(format!("type mismatch, expected `{}`, found `{}`", lhs, rhs)),
    ])
    .with_notes(vec![[
        format!("expected `{}`", lhs),
        format!("   found `{}`", rhs),
    ].join("\n")]])
```

brendanzab (approved):
This looks great! I think part of that was my doing, leading you a bit astray initially, when we could ultimately do it in a much simpler way. Nice to see a simpler solution 😊

fathom/src/surface/elaboration/reporting.rs
wezm: Ahh nice. Yes, that seems better. Will do.

wezm: Done

brendanzab (approved):
Nice, looks really good!

PR #358, opened 2022-05-05
Rename variants in the core language

brendanzab:
Currently I use terminology from [natural deduction](https://ncatlab.org/nlab/show/natural+deduction) to inform the names of the variants in the core language. This has turned out to be a bit intimidating and a source of confusion. The consistency between the names seems to also get in the way of understanding.

In the interest of making it a bit easier to understand and work on Fathom, this PR switches to more familiar names. We can tweak this further down the line, but I hope it helps!

Let me know your thoughts on this.

### DIFF
--- a/fathom/src/core.rs
+++ b/fathom/src/core.rs
@@ -146,13 +146,13 @@ pub enum Term<'arena> {
     Prim(Prim),
 
     /// Constant literals.
-    ConstLit(Const),
-    /// Constant case split.
+    ConstLit(ConstLit),
+    /// Case split on a constant.
     ///
     /// (head_expr, branches, default_expr)
     ConstCase(
         &'arena Term<'arena>,
-        &'arena [(Const, Term<'arena>)],
+        &'arena [(ConstLit, Term<'arena>)],
         Option<&'arena Term<'arena>>,
     ),
 }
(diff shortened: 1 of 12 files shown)


fathom/src/core.rs
```diff
@@ -432,9 +437,9 @@ pub enum UIntStyle {
     Ascii,
 }
 
-/// Constants
+/// Constant literals
 #[derive(Debug, Copy, Clone, PartialOrd)]
-pub enum Const {
+pub enum ConstLit {
```
mikeday, line 442: Why not just `Const`?

wezm:
I think this terminology probably is clearer to someone not well versed in the space. 👍

fathom/src/core.rs
brendanzab: To distinguish it from the `ConstType`s… which don't currently exist 😳… they are currently implemented on the the `Prim`s which makes everything a bit disjointed. I can revert to the old naming though.

brendanzab:
Probably somebody versed in the space would actually call them `Pi`, `Lam`, `App`, and `Sig`, `Struct`, `Proj`, heh. But personally I still kind of like drawing a connection between things related to the different connectives in our type theory (ie. `Fun`, `Record`, `Array`, `Const`, `Format`), and prefer to avoid the Greek letters if possible.

Issue #352, opened 2022-04-28
Sequences of unspecified length


wezm:
OpenType contains sequences with an unspecified length in a number of places. We need a way to handle these.

- Simple glyph flags (`glyf`)
  - The number of flags bytes to read is not specified. Instead bytes need to be read until there's a flag byte for each point. The flags themselves can control what is read after the current byte and may contain a bit indicating that one byte read applies to many points.
- Glyph names in `post`
  - There is an array of Pascal strings and an array that maps glyphs to an index in string array. Reading the string array requires known the maximum index in the mapping array, then read that many (+1) Pascal strings from the string data.

PR #361, opened 2022-05-12
Implement a `repeat_until_end` format.

brendanzab:
This implements a `repeat_until_end` format, along with an `Array` type with no length. We could add more repeat formats later in order to address the needs of #352.

These commits were taken from #310.


wezm (approved):
Looks good 👍 I was expecting it to require more code when I read the PR description.

doc/reference.md
```diff
@@ -355,6 +356,23 @@ of the host array types.
...
+The `repeat_until_end` format repeats parsing the given format until the end of
+the current binary stream is reached:
+
+- `repeat_until_end : Format -> Format`
+
+#### Representation of repeat formats
+
+Because the repeat format does not have a predefined length, it is
+[represented](#format-representations) as a dynamically sized
+[array type](#array-types):
```
mikeday, line 370: Is it possible to access the length once parsed? Do we need to mention that the repeat format doesn't need to be of fixed size?

fathom/src/core/binary.rs
```diff
@@ -183,7 +184,33 @@ impl<'arena, 'env> Context<'arena, 'env> {
...
+                    elems.push(elem);
+                    current_pos = reader.stream_position()?;
+                }
+                Err(err) => match err.kind() {
+                    io::ErrorKind::UnexpectedEof => {
+                        // FIXME: should this be set to the end of the current stream?
+                        reader.seek(SeekFrom::Start(current_pos))?;
+                        return Ok(Arc::new(Value::ArrayLit(elems)));
+                    }
+                    _ => return Err(err),
```
mikeday, line 207: Does this allow trailing random bytes at the end of the stream?

brendanzab: Ahh yeah it does - I was thinking of maybe adding a `repeat_until_end_exact` format.

doc/reference.md
brendanzab: Yeah, I think I'll need to get up to adding sizing to this reference, like how I talk about the host representation types.

Are you thinking the byte length? Or the number of elements in the array?

brendanzab:
Sorry for being hasty in merging this @mikeday! 😳

doc/reference.md
mikeday: I was thinking of number of elements in the array, as the byte length could be determined by following it with a pos.

mikeday:
No problems with the merge, just curious!

PR #362, opened 2022-06-01
Conditional formats

### LINKED ISSUE #339: Conditional formats
It would be useful to have format descriptions that are refine other formats with predicates:

```rust
{ x <- f | e }
```

For example, in OpenType we might see something like:

```rust
{ sfnt_version <- u32be | sfnt_version == 0x00010000 || sfnt_version == "OTTO" },
```

The rules in the core language could look like:

```
  f : Format     x : Repr Format ⊢ e : Bool
──────────────────────────────────────────────
           { x <- f | e } : Format

Repr { x <- f | e } = Repr f
```

When parsing, the guard condition would be checked, and the parse would fail if the condition is false.

Some alternative names for these format descriptions could be:

- conditional formats
- refinement formats
- guard formats

### Predicate preservation

Eventually we could preserve the guard condition in the representation types, using [refinement types](https://github.com/yeslogic/fathom/issues/258):

```diff
-  f : Format     x : Repr Format ⊢ e : Bool
+  f : Format     x : Repr Format ⊢ e : Prop
──────────────────────────────────────────────
           { x <- f | e } : Format


- Repr { x <- f | e } = Repr f
+ Repr { x <- f | e } = { x : Repr f | e }
```

### Syntactic Sugar

We could eventually add some sugar for record formats:

```rust
{
    sfnt_version <- u32be == 0x00010000 || "OTTO",
    ...
}
```

Or perhaps:

```rust
{
    sfnt_version <- u32be if sfnt_version == 0x00010000 || sfnt_version == "OTTO",
    ...
}
```

wezm:
Introduce support for conditional formats. Fixes #339 

This allows values to be verified against a condition. E.g.

```
{ sfnt_version <- { version <- u32be | (u32_eq version 0xffff) } }
```

If the predicate succeeds then the format is read as usual. If it fails this currently results in an error being returned from `read_format`, which in turn is presented as a panic/internal compiler error. I'm opening this PR to get the work reviewed so far and also to discuss how that fail case should be handled. I also need to update the reference and add some more tests when everything is locked down.


brendanzab:
Looks great! We're missing cases in `is_equal` in `core::semantics`, and `unify` in `surface::elaboration::unification`, but things seem to be in order! 👍

brendanzab:
Yeah, it's a bit frustrating that we don't get any errors there :(

wezm:
> Yeah, it's a bit frustrating that we don't get any errors there :(

What do you think of this as a workaround for that?

```rust
    #[test]
    fn value_has_unify_and_is_equal_impls() {
        let value = Arc::new(Value::ConstLit(Const::Bool(false)));

        // This test exists in order to cause a test failure when `Value` is changed. If this test
        // has failed and you have added a new variant to Value it is a prompt to ensure that
        // variant is handled in:
        //
        // - surface::elaboration::Context::unify
        // - core::semantics::is_equal
        //
        // NOTE: Only update the match below when you've updated the above functions.
        match value.as_ref() {
            Value::Stuck(_, _) => {}
            Value::Universe => {}
            Value::FunType(_, _, _) => {}
            Value::FunLit(_, _) => {}
            Value::RecordType(_, _) => {}
            Value::RecordLit(_, _) => {}
            Value::ArrayLit(_) => {}
            Value::FormatRecord(_, _) => {}
            Value::FormatCond(_, _, _) => {}
            Value::FormatOverlap(_, _) => {}
            Value::ConstLit(_) => {}
        }
    }
```

brendanzab:
Oooh, I like that. Clever idea!

doc/reference.md
```diff
@@ -272,6 +273,26 @@ Some examples are as follows:
...
+
+```fathom
+{ major_version <- { version <- u16be | (u16_lte version 2) } }
+```
+
+#### Representation of conditional formats
+
+The [representation](#format-representations) of a conditional format is the same
+as the format without the condition when the condition evaluates to `true`. If
+the condition is `false` then it's equivalent to the [fail format](#fail-format).
```
brendanzab, line 294: I'm not sure this is entirely accurate… this kind of implies something like `Repr format | Void`? Where as we want something more like:

| format                    | `Repr` format         |
| ------------------ | ------------------- |
| `{ x <- format \| cond }` | `Repr format` |

Not sure how to express this better in words, but it might be more accurate to say that “the representation of a conditional format is the same as the representation of the format guarded by the predicate”. It's a bit clunky to describe in words as we don't really have a name for what we're talking about. 😕

wezm: > “the representation of a conditional format is the same as the representation of the format guarded by the predicate”

Happy to change it to this. It's basically what I was was trying to say 😅

brendanzab (approved):
Looks good! 👍

PR #363, opened 2022-06-02
Make read errors slightly better

wezm:
As discussed this is an attempt to make the read errors a bit nicer.

Some of the copy probably still has room for improvement/expanding. Also I wasn't sure about `InvalidFormat`. Is that an error that can be readily user triggered or is not supposed to be encountered in practice? If that latter then it can probably be moved to an `UnexpectedError`, at least for now.

### DIFF
--- a/fathom/src/core/binary.rs
+++ b/fathom/src/core/binary.rs
@@ -2,6 +2,7 @@
 
 use std::collections::HashMap;
 use std::convert::TryFrom;
+use std::fmt;
 use std::slice::SliceIndex;
 use std::sync::Arc;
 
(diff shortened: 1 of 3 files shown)


fathom/src/core/binary.rs
```diff
@@ -375,3 +385,23 @@ read_multibyte_prim!(read_f32le, from_le_bytes, f32);
...
+    }
+}
+
+impl std::error::Error for ReadError {}
+
+impl From<io::Error> for ReadError {
+    fn from(err: io::Error) -> ReadError {
+        ReadError::Io(err)
+    }
+}
```
brendanzab, line 407: Wondering if these impls should be grouped with `ReadError`. 🤔

brendanzab (approved):
This looks like a nice improvement. I can rebase my other PR on top of this if that helps?

brendanzab:
Yeah, I think `InvalidFormat` could probably be moved to `UnexpectedError`.

fathom/src/core/binary.rs
wezm: As in move them up to the struct or move the struct down to the impls?

brendanzab: Maybe grouped with the enum just to keep things in one place.

wezm:
Finally got around to updating this if you wanted to give it a quick once-over @brendanzab.

fathom/src/driver.rs
```diff
@@ -340,4 +344,54 @@ impl<'surface, 'core> Driver<'surface, 'core> {
...
+        // Use the currently set codespan configuration
+        let term_config = self.codespan_config.clone();
+
+        let diagnostic = Diagnostic::from(err);
+        let mut writer = BufferedStandardStream::stderr(if atty::is(atty::Stream::Stderr) {
+            ColorChoice::Auto
+        } else {
+            ColorChoice::Never
+        });
+        let dummy_files = SimpleFiles::<String, String>::new();
```
brendanzab, line 358: I think I use the existing `diagnostic_writer` and file DB in #338. Not sure what is best, but maybe we can get this merged first as it has the friendlier errors! 👍

brendanzab (approved):
Looks nice! 

Looking forward to not having the tests break when `RUST_BACKTRACE=full` is set.

PR #364, opened 2022-06-02
Fill out more of the OpenType format

wezm:
Use some of the recent features to enable more OpenType parsing:

- glyph id array in `cmap` sub-table format 4
- `hmtx` left side bearings
- `maxp` version 1.0 data
- `name` version 1 data
  - I wasn't able to find a font to test this one. Also fonttools doesn't support it so I can't grep through the `ttx` files of the large font collection if have to easily to find a font either.


formats/opentype.fathom
```diff
@@ -418,9 +418,8 @@ let cmap_subtable_format4 = fun (platform : Repr platform_id) => {
     id_delta <- array16 seg_count s16be,
     /// Offsets into glyphIdArray or 0
     id_range_offsets <- array16 seg_count u16be,
-    // TODO: Needs length limiting formats
-    // /// Glyph index array (arbitrary length)
-    // glyph_id_array[ ]  <- u16be,
+    /// Glyph index array (arbitrary length)
+    glyph_id_array <- repeat_until_end u16be,
```
brendanzab, line 422: I'm guessing without #310 this just goes until the end of the binary data?

formats/opentype.fathom
```diff
@@ -823,6 +822,50 @@ let htmx_table = fun (number_of_long_horizontal_metrics : U16) => fun (num_glyph
...
+    /// Maximum number of components referenced at “top level” of all
+    /// composite glyphs.
+    max_component_elements <- u16be,
+    /// Maximum levels of recursion used when constructing compound glyphs.
+    ///
+    /// | Value | Meaning                                           |
+    /// | ----- | ------------------------------------------------- |
+    /// | `0`   | the font only contains simple glyphs              |
+    /// | `1`   | compound glyphs only contain simple glyphs (there are no components within components) |
+    max_component_depth <- { depth <- u16be | u16_lte depth 16 }, // where max_component_depth in 0..16
```
brendanzab, line 865: We really do need some sugar for this haha 😅

brendanzab (approved):
Looks great 👍

formats/opentype.fathom
wezm: Oh right… I was thinking that it stopped at the end of the table… perhaps this change was premature.

formats/opentype.fathom
brendanzab: Sugar has been added in #370! 🍭

wezm: Oops I forgot about this PR. I'll give it a touch up now.

wezm:
I've updated the PR but using `opentype.fathom` is now giving this error:

```
        error: escaping rigid variable
             ┌─ formats/opentype.fathom:1881:33
             │
        1881 │                 maxp <- deref _ maxp.link,
             │                                 ^^^^^^^^^

[✂ snip many more errors after this one]
```

I'm not sure why my changes are triggering it. Any ideas @brendanzab?

brendanzab:
Oh no! I think I was running into a similar error originally, which is why I’d annotated the `deref`s prior to #366. If all else fails we could try going back to that. 🤔

wezm:
> Oh no! I think I was running into a similar error originally, which is why I’d annotated the `deref`s prior to #366. If all else fails we could try going back to that. 🤔

Ahh cool. Adding the annotations back for maxp fixes it.

brendanzab:
Weird… something spooky must be going on. Will need to look into it I guess. `left_side_bearings` look nice now at least!

brendanzab (approved):

Issue #367, opened 2022-06-07
Let formats


brendanzab:
In the OpenType we sometimes have to employ records as a way to store intermediate values during parsing. For example:

```
htmx <- required_table "hmtx" {
    hhea <- deref _ hhea.link,
    maxp <- deref _ maxp.link,
    table <- htmx_table
        hhea.number_of_long_horizontal_metrics
        maxp.num_glyphs,
},
```

This is not ideal as this results in a copy of `hhea` and `maxp` appearing in the resulting data structures after parsing. For example:

```
4752 = [
    {
        hhea = {
            major_version = 1,
            ⋮
            number_of_long_horizontal_metrics = 1,
        },
        maxp = { version = 20480, num_glyphs = 100 },
        table = {
            h_metrics = [ { advance_width = 1500, left_side_bearing = 300 } ],
        },
    },
]
```

It would be nice however to be able to write something like:

```
htmx <- required_table "hmtx" (
    let hhea <- deref _ hhea.link;
    let maxp <- deref _ maxp.link;
    htmx_table
        hhea.number_of_long_horizontal_metrics
        maxp.num_glyphs,
),
```

This would employ a new format of the form: `let x <- f₁; f₂`, which allows a format to be parsed, with the result added to the environment, and then a subsequent format will be parsed in that environment. This would result in a data structure that looks something like:

```
4752 = [
    {
        h_metrics = [ { advance_width = 1500, left_side_bearing = 300 } ],
    },
]
```

This also has the tantalizing possibility of being further improved to look something like:

```
htmx <- required_table "hmtx" (
    htmx_table
        (deref _ hhea.link).number_of_long_horizontal_metrics
        (deref _ maxp.link).num_glyphs,
),
```

## Rough Specification

The typing and parsing rules for this format could look something like:

```
  f₁ : Format     x : Repr Format ⊢ f₂ : Format
─────────────────────────────────────────────────
           let x <- f₁; f₂ : Format


  s .. s' : f₁ ⟹ e₁   x : Repr f = e₁ ⊢ s' .. s'' : f₁ ⟹ e₂
────────────────────────────────────────────────────────────────
          s .. s'' : (let x <- f₁; f₂) ⟹ e₂
```

Alas, difficulty arises when attempting to define a host representation for this format.

```
Repr (let x <- f₁; f₂ x) = Repr (f₂ x)
                                    ^ where is `x : Repr f₁` bound?
```

I'm not sure yet how to resolve this.

[...]

PR #369, opened 2022-06-08
Add GDEF definition to OpenType format

wezm:
Turns out there is _a lot_ in `GDEF`. The weirdest/most interesting bit is probably `u16_div_ceil` where I need the ceiling of a division... so I implemented that with just integer ops. Unfortunately I've been unable to find a single font that exercises that path.


formats/opentype.fathom
```diff
@@ -185,6 +186,197 @@ let language_id =
...
+/// # Device and VariationIndex Tables
+///
+/// Device tables and VariationIndex tables are used to provide adjustments to font-unit values in
+/// GPOS, JSTF, GDEF or BASE tables, such as the X and Y coordinates of an attachment anchor
+/// position.
+///
+/// Curiously the table has two interpretations. The second interprtation appears to be have been
+/// tacked on for variable fonts. The gist being that if the delta format is 0x8000 then the table
+/// is a VariationIndex table, which names the fields differently and does not contain a delta
+/// value array. E.g.
```
brendanzab, line 320: Ahhhh, I get it now! It's not particularly elegant, but this does seems like a place where an `overlap` format could be used:

```
overlap {
    // Initial pass to figure out the table format
    init <- {
        _skipped <- array8 4 u8,
        table_format <- u16be,
    },
    // Device and VariationIndex Tables
    table <- match init.table_format {
        0x0001 => device_table,
        0x0002 => device_table,
        0x0003 => device_table,
        0x8000 => variation_index_table,
        0x7FFC => unknown_table, // not sure if this was meant to be “anything up to`0x7FFC`”?
        _ => fail,
    },
}
```

formats/opentype.fathom
```diff
@@ -185,6 +186,197 @@ let language_id =
...
+    {
+        /// Smallest size to correct, in ppem
+        start_size <- u16be,
+        /// Largest size to correct, in ppem
+        end_size <- u16be,
+        /// Format of deltaValue array data
+        delta_format <- u16be,
+        /// Array of compressed data
+        _delta_bits <- succeed _ (delta_bits delta_format (num_sizes start_size end_size)),
+        delta_values <- array16 (u16_div_ceil _delta_bits 16) u16be,
```
brendanzab, line 376: Could maybe use a `let` here:

```
        /// Array of compressed data
        delta_values <- 
            let delta_bits = delta_bits delta_format (num_sizes start_size end_size);
            array16 (u16_div_ceil delta_bits 16) u16be,
```

formats/opentype.fathom
```diff
@@ -185,6 +186,197 @@ let language_id =
...
+///
+/// - [Microsoft's OpenType Spec: Class Definition Table](https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#class-definition-table)
+let class_def = {
+    /// Format identifier
+    class_format <- u16be,
+    /// Format specific data
+    data <- match class_format {
+        1 => class_def_format_1,
+        2 => class_def_format_2,
+        _ => fail,
```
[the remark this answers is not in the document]
wezm: This is a bit of an open question in my mind, at least partially informed by the use case. I'm picturing two use-cases: exploring a font file, and parsing and using that data in a font file. I suppose I've been thinking more along the lines of the latter, which might actually be wrong for where we're at.

Where I've used `fail` in this PR there is no data to fall back on. For example, If someone was relying on getting class def information we'd be unable to provide it if it used a format we were unaware of. Presumably the alternative is return something in place of failing, do you have thoughts on what that could be?

In other places like the handling of the `GDEF` minor version we fall back on the newest version we know about as each of the revisions is additive so we can read up to the version we know.

formats/opentype.fathom
wezm: >         0x7FFC => unknown_table, // not sure if this was meant to be “anything up to`0x7FFC`”?

From the comment in the spec I think they mean that each of the bits in `0x7FFC` (as opposed to that particular value) are reserved for future use (and should be set to zero):

> For future use — set to 0

formats/opentype.fathom
wezm: That works!? 🤯 

I've been operating on the assumption that you couldn't use let there and was evidently quick mistaken about #367

wezm: So the only reason that `succeed` is used here is because that value needs to be used by several fields?

https://github.com/yeslogic/fathom/blob/811673a3833c66942e3282ad1e8c396c05d0ab1b/formats/opentype.fathom#L400

brendanzab: Yeah that's right! Introducing shared definitions into a record (which is useful if you don't want to recompute something multiple times, but also just for clarity). I think you can actually kind of think of `succeed` as a bit like the `pure` function in Haskell (aka. `return`). This embeds a ‘pure’ computation in the middle of some binary data.

I've been doing a bit of pondering about this today, thinking some different approaches. A simple way in the short term could be to have something like:

```
let seg_count = u16_div seg_count_x2 2,
```

which would then be elaborated to:

```
seg_count <- succeed _ (u16_div seg_count_x2 2),
```

This still means that it shows up in the resulting record however, which could be a bit of a pain. Still don't have a good solution for this alas.

> I've been operating on the assumption that you couldn't use let there and was evidently quite mistaken about https://github.com/yeslogic/fathom/issues/367

Ohh this might have been a miscommunication. The ‘let format’ in that issue could be seen as ‘effectful’ - in that it lets you do parsing as well. In this case there's no need to do any parsing, so a regular `let` is ok.

formats/opentype.fathom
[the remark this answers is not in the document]
brendanzab: I think the way I've been handling it is with the `unknown_table` format which is just a empty record format that always succeeds. 

```
/// # Unknown table format
///
/// This is a placeholder for a table that has an unknown identifier (due to the
/// file conforming to a newer version of the specification), or for a table has
/// not yet been implemented.
let unknown_table : Format = {};
```

It used to be used as a fallback case to a match arm, before we switched to our current approach for font tables. This meant that the match arm would succeed if a new version was introduced that we didn't recognise, but no data would be read. Unfortunately this approach needs to be used with care… like you need to use it at the end of a substream, otherwise the format could continue to be read afterwards. Though I think this would only really occur in a poorly designed format I guess.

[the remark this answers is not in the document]
brendanzab: Ahh, we currently use it for the cmap tables:

https://github.com/yeslogic/fathom/blob/811673a3833c66942e3282ad1e8c396c05d0ab1b/formats/opentype.fathom#L565-L583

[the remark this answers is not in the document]
brendanzab: I’ve had a bit of a think about it, and don't think this is a huge issue right now, so feel free to ignore me. And yeah thanks for the consideration of the different use cases… seems that the desired behavior might be a bit application specific. Food for thought at the very least!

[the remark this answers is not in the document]
wezm: Heh I pondered it too and decided to change it to `unknown_table`:

- Nothing else was using fail.
- The doc comments on `unknown_table` are clear that it's for cases like this.
- We can change it to something like `{ unknown <- {} }` as you proposed in the future to make it show up in the output if we want.

brendanzab (approved):
Awesome, looks good!

PR #370, opened 2022-06-09
Implement `where` sugar for refined fields

### LINKED ISSUE #368: Sugar for guarded fields in record formats
A common pattern in the OpenType specification and binary formats in general is to write something like:

```
let table_directory = fun (file_start : Pos) => {
    sfnt_version <- { version <- u32be | version == 0x00010000 || version == "OTTO" },
    ⋮
};
```

It would be nice to alter the grammar of record formats to allow us write something like this:

```
let table_directory = fun (file_start : Pos) => {
    sfnt_version <- u32be when 
        sfnt_version == 0x00010000 || sfnt_version == "OTTO",
    ⋮
};
```

These guarded fields would be elaborated to conditional formats, so no changes would need to be made in the core language or the binary parser semantics. 

The `when` keyword was chosen rather arbitrarily. `if` would probably be nicer, but I'm not sure we’d be able to fit it into the grammar that easily. I could see some confusion arising if somebody thought that `when` meant that the field is optionally present.

As an aside, this would still be a way off Mike’s dream of being able to write something like:

```
let table_directory = fun (file_start : Pos) => {
    sfnt_version <- u32be == 0x00010000 || "OTTO",
    ⋮
};
```

brendanzab:
This implements some short-hand sugar for refining fields with predicates. Closes #368.

```
{
    magic <- u32be where u32_eq magic "icns",
    ⋮
}
```

These are turned into conditional formats during elaboration. I went with calling them ’field refinements’ as if felt like ‘conditional fields’ could be interpreted as being optional. Another name for them could be ‘field guards’, perhaps.

I also found a bug in the implementation of `Repr` for conditional fields in the process!

### DIFF
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -265,16 +265,16 @@ array8 3 {}
 #### Field refinements
 
 The parsed representations of fields can be refined with boolean predicates
-using the `when` syntax. If the predicate evaluates to `false`, the rest of the
+using the `where` syntax. If the predicate evaluates to `false`, the rest of the
 record format will fail to parse. For example:
 
 ```fathom
 {
-    magic <- u32be when u32_eq magic "icns",
-    //                  ▲      ▲
-    //                  │      └──── `magic` is bound as type `Repr u32be`
-    //                  │
-    //                  └──── `u32_eq magic "icns"` must be of type `Bool`
+    magic <- u32be where u32_eq magic "icns",
+    //                   ▲      ▲
+    //                   │      └──── `magic` is bound as type `Repr u32be`
+    //                   │
+    //                   └──── `u32_eq magic "icns"` must be of type `Bool`
 }
 ```
 
@@ -301,7 +301,7 @@ Some examples are as follows:
 | `{}`                                              | `{}`                                   |
 | `{ x <- f32le, y <- f32le }`                      | `{ x : F32, y : F32 }`                 |
 | `{ len <- u16be, data <- array16 len s8 }`        | `{ len : U16, data : Array16 len S8 }` |
-| `{ magic <- u32be when u32_eq magic "icns" }`     | `{ magic : U32 }`                      |
+| `{ magic <- u32be where u32_eq magic "icns" }`    | `{ magic : U32 }`                      |
 
 ### Conditional formats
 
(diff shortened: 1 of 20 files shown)


doc/reference.md
```diff
@@ -261,6 +262,32 @@ the type of `array8`:
 array8 3 {}
 ```
 
+#### Field refinements
+
+The parsed representations of fields can be refined with boolean predicates
+using the `where` syntax. If the predicate evaluates to `false`, the rest of the
```
wezm, line 268:
```suggestion
using the `when` syntax. If the predicate evaluates to `false`, the rest of the
```

fathom/src/surface.rs
```diff
@@ -198,6 +198,35 @@ impl<'arena> Term<'arena, ByteRange> {
...
+}
+
+/// A field definition in a record literal
+#[derive(Debug, Clone)]
+pub struct ExprField<'arena, Range> {
+    /// Label identifying the field
+    label: (Range, StringId),
+    /// The expression that this field will store
+    expr: &'arena Term<'arena, Range>,
+}
```
wezm: These are a nice improvement to readability 👍

fathom/src/surface/elaboration.rs
```diff
@@ -760,25 +760,24 @@ impl<'interner, 'arena, 'error> Context<'interner, 'arena, 'error> {
...
+    fn report_duplicate_labels<'fields, F>(
         &mut self,
         range: ByteRange,
-        fields: &'fields [((ByteRange, StringId), Term<'a, ByteRange>)],
-    ) -> (
-        &'arena [StringId],
-        impl Iterator<Item = &'fields ((ByteRange, StringId), Term<'a, ByteRange>)>,
-    ) {
+        fields: &'fields [F],
+        get_label: fn(&F) -> (ByteRange, StringId),
```
wezm, line 767: Could this be a trait on `F` instead of passing a function?

Although looking at the closures that are actually passed perhaps it's not worth the ceremony.

wezm (approved):
Overall, looks great.

🚴‍♂️ bike shedding: I feel like `where` reads better than `when`?

doc/reference.md
brendanzab: Yeah I constantly get these mixed up haha… I’ll try `when`

PR #371, opened 2022-06-10
Add surface syntax for computed fields

brendanzab:
This implements some shorthand syntax for adding pure computations into a record format. For example:

```fathom
{
    /// 2 × seg_count.
    seg_count_x2 <- u16be,

    /// Number of contiguous ranges of character codes
    let seg_count = u16_div seg_count_x2 2,
    //  ▲
    //  └─── the value computed for `seg_count` will be
    //       available for use in subsequent fields

    ⋮

    start_code <- array16 seg_count u16be,
    id_delta <- array16 seg_count s16be,
    //                  ▲
    //                  └──── `seg_count` is used here
}
```

This is translated to a `succeed` format during elaboration.

### DIFF
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -288,20 +288,77 @@ format. For example, the above format is equivalent to:
 }
 ```
 
+#### Computed fields
+
+Sometimes it is useful to embed a pure computation (that does not perform any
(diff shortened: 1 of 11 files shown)


wezm (approved):
Looks great 👍

PR #373, opened 2022-06-15
Add GPOS definition to OpenType format

wezm:
As per title. The build will fail as I didn't update the `formats/data/opentype/EncodeSans[wdth,wght].snap` snapshot… after these changes it grows to 6.5Mb, which is perhaps a little excessive. <del>I'm not certain it's worth keeping the font in the repo 🤔 </del> It probably should be removed as it slows the tests down quite a bit.


formats/opentype.fathom
```diff
@@ -800,7 +800,7 @@ let single_substitution = {
     /// Format identifier
     subst_format <- u16be,
     /// Coverage table
-    coverage <- offset16 table_start coverage_table,
+    coverage <- offset16 table_start coverage_table, // TODO move
```
wezm, line 803: Hmm need to address this.

formats/opentype.fathom
```diff
@@ -1036,6 +1038,527 @@ let extension_substitution = (
...
+    let X_PLACEMENT_DEVICE : U16 = 0x0010;
+    let Y_PLACEMENT_DEVICE : U16 = 0x0020;
+    let X_ADVANCE_DEVICE : U16 = 0x0040;
+    let Y_ADVANCE_DEVICE : U16 = 0x0080;
+
+    let optional_field = fun (field : U16) => fun (format : Format) =>
+        match (u16_neq (u16_and flags field) 0) {
+            true => format,
+            false => empty,
+        };
```
mikeday, line 1099: This is nice!

formats/opentype.fathom
```diff
@@ -1036,6 +1038,527 @@ let extension_substitution = (
...
+        /// — may be NULL.
+        y_adv_device_offset <- optional_field Y_ADVANCE_DEVICE (offset16 table_start device_or_variation_index_table),
+    }
+);
+
+/// A value record that is `empty` if flags is 0
+let optional_value_record = fun (table_start : Pos) => fun (flags : U16) => match (u16_eq flags 0) {
+    true => empty,
+    false => value_record table_start flags
+};
```
mikeday, line 1137: Won't `value_record` already be empty if `flags` is zero?

wezm: Yep. I initially was going to just use `value_record` but doing that results in something like the following in the output:

```
{
    x_placement <- empty,
    y_placement <- empty,
    x_advance <- empty,
    y_advance <- empty,
    x_pla_device_offset <- empty,
    y_pla_device_offset <- empty,
    x_adv_device_offset <- empty,
    y_adv_device_offset <- empty,
}
```

instead of just `empty`. I thought having it named `optional_value_record` at the call-site was useful for clarity too.

mikeday (approved):
That looks great! Being able to share the definitions across GSUB and GPOS is perfect.

brendanzab (approved):
Looks good, to the extent I understand this part of OpenType at least!

PR #375, opened 2022-06-21
Provide more information in SetOffsetOutsideBuffer error

wezm:
I ran into this error when I was working on adding `GPOS` to the OpenType definition and wished there was more info in it. I've added the bad offset to the error, which may or may not actually be useful. I've also moved it out of the `bug` reporting branch when reporting errors since it can be triggered with bad data or bad format description.

### DIFF
--- a/fathom/src/core/binary.rs
+++ b/fathom/src/core/binary.rs
@@ -18,7 +18,7 @@ pub enum ReadError {
     UnwrappedNone,
     ReadFailFormat,
     CondFailure,
-    SetOffsetOutsideBuffer,
+    SetOffsetOutsideBuffer(usize),
     UnexpectedEndOfBuffer,
     PositionOverflow,
 }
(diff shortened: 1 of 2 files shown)


fathom/src/core/binary.rs
```diff
@@ -165,18 +165,16 @@ impl<'data> BufferReader<'data> {
...
         (relative_offset <= self.buffer.remaining_len())
             .then(|| self.relative_offset = relative_offset)
-            .ok_or(ReadError::SetOffsetOutsideBuffer)
     }
 
     /// Set the offset of the reader relative to the start position.
     pub fn set_offset(&mut self, offset: usize) -> Result<(), ReadError> {
-        match usize::checked_sub(offset, self.buffer.start_offset) {
-            Some(relative_offset) => self.set_relative_offset(relative_offset),
-            None => Err(ReadError::SetOffsetOutsideBuffer),
```
brendanzab, line 178: Maybe could change to `ReadError::SetOffsetBeforeStartOfBuffer { offset }`. And `ReadError::SetOffsetAfterEndOfBuffer { offset: Option<T> }` (the `Option` to handle the exceedingly unlikely chance of an overflow). Then both `set_relative_offset` and `set_offset` can return results, with better errors.

PR #376, opened 2022-06-24
Out of order definitions

wezm:
This PR allows you to define items in a natural order, instead of requiring item references to be defined first.


fathom/src/surface.rs
```diff
@@ -42,6 +47,383 @@ impl<'arena> Module<'arena, ByteRange> {
...
+        Some(ScopeItem::Name(name)) => {
+            visited.insert(*name);
+        }
+        Some(ScopeItem::Pending) | None => {}
+    }
+}
+
+#[derive(Copy, Clone)]
+enum ScopeItem<'a, 'arena> {
+    Definition(&'a Item<'arena, ByteRange>),
```
wezm, line 171: I don't think this actually needs to carry the Item. I'll try removing it, which should tidy things up a bit.

brendanzab:
Looking good! Would it be possible to move the module order stuff into `elaboration`? Maybe it could go into a private submodule, maybe `elaboration::item_order` or something? This would then be called in `elab_module`. That means we we have less mixing of elaboration related stuff in with the top-level `surface` module, which is more just related to defining the AST, and means that the driver does not have to interact with it either.

brendanzab:
Still having a go at trying to understand how the algorithm works – sorry for my slowness!

wezm:
> Still having a go at trying to understand how the algorithm works – sorry for my slowness!

Hmm perhaps I should add a comment documenting the approach

wezm:
Basically it's:

- For each item recurse into all the terms it contains and note the dependencies (that aren't local names).
- Then do a depth first traversal of an item's dependencies and their dependencies pushing the indexes of items as we hit leaves in order to produce a list of item indexes describing the order in which they should be elaborated.
  - At the same time keep a stack of names in the traversal so that we can detect when we've re-entered a name that we're in the process of visiting and report a cycle.

wezm:
Pushed an update to move the code into a sub-module. That was a good suggestion, thank you.

fathom/src/surface/elaboration.rs
```diff
@@ -981,14 +983,20 @@ impl<'interner, 'arena, 'error> Context<'interner, 'arena, 'error> {
     }
 
     /// Elaborate a module
-    pub fn elab_module(&mut self, surface_module: &Module<'_, ByteRange>) -> core::Module<'arena> {
+    pub fn elab_module(
+        &mut self,
+        surface_module: &Module<'_, ByteRange>,
+    ) -> Result<core::Module<'arena>, Vec<ModuleOrderMessage>> {
```
brendanzab, line 989: I wonder if we could also add the messages to the elaboration context, to avoid needing to deal with them separately in the driver.

wezm: Oh of course, this makes a lot of sense now that it's been moved into `elab_module`.

wezm: I've updated the PR again with this change. Thanks again for the suggestions. It seems to have worked out pretty cleanly.

tests/cmd/fathom-elab.md
```diff
@@ -144,3 +144,22 @@ error: duplicate labels found in record
...
+$ fathom elab --module tests/fail/elaboration/item-cycle.fathom
+? failed
+error: cycle detected
+ = first → second → third → first
+
+error: cannot find `first` in scope
+  ┌─ tests/fail/elaboration/item-cycle.fathom:5:13
+  │
+5 │ def third = first;
+  │             ^^^^^ unbound name
```
brendanzab, line 162: Curious why the body of this item is getting elaborated 🤔 - is the index of the item getting returned from `determine_order`?

wezm: Yes it does: `visit_item` recurses into `first`s dependencies and encounters the cycle so it stops recursing. When that returns the index of `first` is added to the output since it's assumed that all its dependencies have been added before it.

https://github.com/yeslogic/fathom/pull/376/files#diff-936b5db84b3d389fc600527214d8189638f749750ff83bf3702c6379c19e7cadR130

Should it omit `first`s index if errors are detected visiting its dependencies?

brendanzab: Yeah I think so - if the programmer wrote `first` they might get a bit frustrated if this was getting reported as being unbound. Not sure how hard it would be with the current algorithm, though!

wezm: I think it's a pretty easy change. I'll give it a go

wezm: Ok I was wrong about it being easy. I went around in circles for a bit but I think came up with a solution that handles it reasonably nicely in the end.

mikeday: Went around in circles you say,

wezm: Indeed, but that connection was accidental.

brendanzab (approved):
Awesome work - the OpenType file looks so much more readable now. I think maybe there _might_ be some more simplifications possible in the order analysis, but it would be great to get this merged so that we can start using it. Thanks so much for this!

fathom/src/surface/elaboration.rs
```diff
@@ -982,13 +983,15 @@ impl<'interner, 'arena, 'error> Context<'interner, 'arena, 'error> {
 
     /// Elaborate a module
     pub fn elab_module(&mut self, surface_module: &Module<'_, ByteRange>) -> core::Module<'arena> {
+        let elab_order = order::elaboration_order(self, surface_module);
         let universe = Arc::new(Value::Universe);
+        // TODO: Could the Module keep track of this as items are added to it
         let num_items = (surface_module.items.iter())
             .filter(|item| matches!(item, Item::Definition { .. }))
             .count();
         let mut items = SliceVec::new(self.scope, num_items);
```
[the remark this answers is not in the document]
wezm: Excellent point. I've made that change.

PR #378, opened 2022-07-12
Sugar for comparison operators

wezm:


brendanzab (approved):
Great! Might need some reference docs for these operators at some stage too.

PR #379, opened 2022-07-19
Avoid allocating Vec every time we add to reference cache

wezm:
Noticed this doing some other research. Prior to this change the `Vec` was allocated every time, even if the entry was already populated.


brendanzab (approved):
Nice, thanks!

PR #380, opened 2022-07-25
Distill `core::Term::FunType` to `surface::Term::Arrow` if `input_name` is not bound in `output_type`

Kmeakin:
Results in more readable pretty-printing when printing non-dependent function types (eg `fun (_: Type) -> Type` is now rendered as `Type -> Type`


brendanzab:
This is really awesome! I was a bit stumped on how to do this nicely - thanks for looking into it!

PR #381, opened 2022-07-28
Track source locations in core::Term and Values

wezm:
This PR adds source location information to `core::Term` and `Value` so that this information can flow from the surface language through the core language and back out. This allows some errors to be much more helpful.

This is a large change but I tried to keep the commits focussed on a single `Term` or `Value` variant so the history should be quite readable. There are a number of `FIXME`, `TODO` comments and a few references to a `Span::fixme` method left. Most of them are questions that I want to discuss to determine the right approach to resolve them.

I initially added spans to `Value` directly but this made it difficult when reusing a shared value instance (since they are ref-counted). For example, when encountering a `fail` format the value is the shared fail instance in the environment but the span is wherever this particular use of fail occurred. As a result of wanting to be able to provide a decent error in situations like this I refactored the handling of `Values` to instead have a wrapper type (`SpanValue`) that contains a `Span` and a `Value`. This way the span is independent of the `Value`. This worked out well for the `fail` case, allowing the location in the format to be included in the error.

There's a similar case that I would like to work but have not worked out how to do it—I did add a test though. When you unwrap none the error is:

```
error: unwrapped none
 = option_unwrap was called on a none value.
```

I'd like that to be able to point at where the unwrap occurred but got a bit lost in `step` macro.

Another thing I'm unsure about is things like this (from `synth`):

```rust
            Term::RecordLiteral(range, expr_fields) => {
                ⋮
                let types = Telescope::new(self.rigid_env.exprs.clone(), types.into());

                (
                    core::Term::RecordLit(range.into(), labels, exprs.into()),
                    SpanValue::empty(Arc::new(Value::RecordType(labels, types))),
                )
            }
```

where we return a term an its type. Should the type use the same range as the term? This question accounts for a bunch of the TODOs/FIXMEs I have in different places.

### DIFF
--- a/fathom/src/core/binary.rs
+++ b/fathom/src/core/binary.rs
@@ -339,9 +339,9 @@ impl<'arena, 'env, 'data> Context<'arena, 'env, 'data> {
             Value::Stuck(span, Head::RigidVar(_), _)
             | Value::Stuck(span, Head::FlexibleVar(_), _)
             | Value::Universe(span)
-            | Value::FunType(span, _, _, _) => Err(ReadError::InvalidFormat(*span)),
-            Value::FunLit(_, _)
-            | Value::RecordType(_, _)
+            | Value::FunType(span, _, _, _)
+            | Value::FunLit(span, _, _) => Err(ReadError::InvalidFormat(*span)),
+            Value::RecordType(_, _)
             | Value::RecordLit(_, _)
             | Value::ArrayLit(_)
             | Value::ConstLit(_) => Err(ReadError::InvalidFormat(Span::fixme())),
--- a/fathom/src/core/binary.rs
+++ b/fathom/src/core/binary.rs
@@ -338,7 +338,7 @@ impl<'arena, 'env, 'data> Context<'arena, 'env, 'data> {
                     }
                 }
             }
-            Value::FormatOverlap(span, labels, formats) => {
+            Value::FormatOverlap(labels, formats) => {
                 let mut max_relative_offset = reader.relative_offset();
 
                 let mut formats = formats.clone();
@@ -361,7 +361,10 @@ impl<'arena, 'env, 'data> Context<'arena, 'env, 'data> {
                 // reached in loop above.
                 reader.set_relative_offset(max_relative_offset).unwrap();
 
-                Ok(SpanValue(*span, Arc::new(Value::RecordLit(labels, exprs))))
+                Ok(SpanValue(
+                    format_span,
+                    Arc::new(Value::RecordLit(labels, exprs)),
+                ))
             }
 
             Value::Stuck(Head::RigidVar(_), _) | Value::Stuck(Head::FlexibleVar(_), _) => {
--- a/fathom/src/core/binary.rs
+++ b/fathom/src/core/binary.rs
@@ -20,8 +20,7 @@ pub enum ReadError<'arena> {
     UnwrappedNone(Span),
     ReadFailFormat(Span),
     CondFailure(Span, ArcValue<'arena>),
-    BufferError(BufferError),
-    BufferErrorWithSpan(Span, BufferError),
+    BufferError(Span, BufferError),
 }
 
 impl<'arena> fmt::Display for ReadError<'arena> {
@@ -33,8 +32,7 @@ impl<'arena> fmt::Display for ReadError<'arena> {
             ReadError::UnknownItem => f.write_str("unknown item"),
             ReadError::ReadFailFormat(_) => f.write_str("read a fail format"),
             ReadError::CondFailure(_, _) => f.write_str("conditional format failed"),
-            ReadError::BufferError(err) => fmt::Display::fmt(&err, f),
-            ReadError::BufferErrorWithSpan(_span, err) => fmt::Display::fmt(&err, f),
+            ReadError::BufferError(_, err) => fmt::Display::fmt(&err, f),
         }
     }
 }
@@ -43,7 +41,7 @@ impl<'arena> std::error::Error for ReadError<'arena> {}
 
 impl<'arena> From<BufferError> for ReadError<'arena> {
     fn from(err: BufferError) -> Self {
-        ReadError::BufferError(err)
+        ReadError::BufferError(Span::Empty, err)
     }
 }
 
@@ -228,7 +226,7 @@ pub enum BufferError {
 
 impl BufferError {
     fn with_span<'arena>(self, span: Span) -> ReadError<'arena> {
-        ReadError::BufferErrorWithSpan(span, self)
+        ReadError::BufferError(span, self)
     }
 }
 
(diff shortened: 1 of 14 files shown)


wezm:
In the latter commits I reworked how `BufferError` is handled to allow errors like this to be generated:

```
error: unexpected end of buffer
   ┌─ formats/opentype.fathom:34:18
   │
34 │         magic <- u32be,
   │                  ^^^^^
   │
   = The end of the buffer was reached before all data could be read.
```

fathom/src/surface/distillation.rs
```diff
@@ -86,13 +87,14 @@ impl<'interner, 'arena, 'env> Context<'interner, 'arena, 'env> {
         &mut self,
         number: T,
         style: UIntStyle,
-    ) -> Term<'arena, ()> {
+        span: Span,
+    ) -> Term<'arena, Span> {
```
brendanzab, line 91: We might be able to unparameterise the surface terms at this point 🤔

brendanzab (commented):
Just posting some thoughts before I forget, sorry for the slowness

fathom/src/core/semantics.rs
```diff
@@ -8,11 +8,35 @@ use std::sync::Arc;
 use crate::alloc::SliceVec;
 use crate::core::{Const, EntryInfo, Prim, Term, UIntStyle};
 use crate::env::{EnvLen, GlobalVar, SharedEnv, SliceEnv};
+use crate::source::Span;
 use crate::StringId;
 
 /// Atomically reference counted values. We use reference counting to increase
 /// the amount of sharing we can achieve during evaluation.
-pub type ArcValue<'arena> = Arc<Value<'arena>>;
+pub type ArcValue<'arena> = SpanValue<'arena>;
```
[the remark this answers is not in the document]
wezm: > Is there any reason where you chose This, as opposed to `Arc<SpanValue<'arena>>`?

Yes, my thinking was that this allows `Values` to be reused/shared but if we put the span inside the Arc then they can only shared if they are from the same source location. This came up when adding a span for `fail`. Since `fail` refers to the same value in the environment I needed a way to attach source location to it when it was encountered. Does that make sense? It's also possible that I've missed something.

fathom/src/core/semantics.rs
```diff
@@ -8,11 +8,35 @@ use std::sync::Arc;
...
+
+#[derive(Debug, Clone)]
+pub struct SpanValue<'arena>(pub Span, pub Arc<Value<'arena>>);
```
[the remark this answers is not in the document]
wezm: Yep that seems reasonable.

fathom/src/surface/distillation.rs
wezm: I started doing this but I'm not sure I like the end result. Since we need to be able to produce `Terms` with `Spans` I assume that the `Range` parameter would be replaced with `Span`. However that means that all the places where the parameter was `ByteRange` now accept/need to handle `Span::Empty`, which feels less accurate (Since Term<ByteRange> always has a file id and range). I think for now it might be worth leaving this as-is. Happy to debate the pros and cons though.

fathom/src/core/semantics.rs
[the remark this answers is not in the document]
wezm: Pushed some more commits that implement this.

fathom/src/core/binary.rs
```diff
@@ -3,49 +3,47 @@
...
-    ReadFailFormat,
-    CondFailure,
-    SetOffsetBeforeStartOfBuffer { offset: usize },
-    SetOffsetAfterEndOfBuffer { offset: Option<usize> },
-    UnexpectedEndOfBuffer,
-    PositionOverflow,
+    UnwrappedNone(Span),
+    ReadFailFormat(Span),
+    CondFailure(Span, ArcValue<'arena>),
+    BufferError(Span, BufferError),
```
brendanzab, line 23: I like splitting this out! 👍

fathom/src/surface/distillation.rs
brendanzab: Tbh I'm a little uncomfortable returning `Spans` here as I kind of consider distilled code to be generated. Maybe in the future we could add some ‘generated span’ variant that points back to the original source span. 🤔

fathom/src/surface/elaboration.rs
```diff
@@ -118,27 +118,27 @@ impl<'arena> RigidEnv<'arena> {
...
+        const U64_TYPE: Term<'_> = Term::Prim(Span::Empty, U64Type);
+        const S8_TYPE: Term<'_> = Term::Prim(Span::Empty, S8Type);
+        const S16_TYPE: Term<'_> = Term::Prim(Span::Empty, S16Type);
+        const S32_TYPE: Term<'_> = Term::Prim(Span::Empty, S32Type);
+        const S64_TYPE: Term<'_> = Term::Prim(Span::Empty, S64Type);
+        const ARRAY8_TYPE: Term<'_> = Term::Prim(Span::Empty, Array8Type);
+        const ARRAY16_TYPE: Term<'_> = Term::Prim(Span::Empty, Array16Type);
+        const ARRAY32_TYPE: Term<'_> = Term::Prim(Span::Empty, Array32Type);
+        const ARRAY64_TYPE: Term<'_> = Term::Prim(Span::Empty, Array64Type);
+        const POS_TYPE: Term<'_> = Term::Prim(Span::Empty, PosType);
```
brendanzab, line 141: Eventually would be cool if these could point to some place in a `prims` module.

fathom/src/surface/elaboration.rs
```diff
@@ -2177,6 +2445,7 @@ impl<'interner, 'arena, 'error> Context<'interner, 'arena, 'error> {
                         );
 
                         core::Term::Let(
+                            (&range).into(), // FIXME: is this the right range to use here?
```
brendanzab, line 2448: This is tricky, as it's sort of ‘generated code’… so it’s a little hard to give a precise range… it should probably at least be the range of the match arm… currently this is just the pattern 🤔

fathom/src/core/semantics.rs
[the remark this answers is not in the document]
brendanzab: Maybe `fail` could take a source location as part of the primitive? 🤔 eg.

```
SourceSpan : Type
fail : SourceSpan -> Format
```

Oh wait, this would only work for `reported_error`s as those are only even created by the elaborator. You need to make fail formats as a programmer… if we had glued eval we might be able to take it from glued thing. Hmm, need to think. Anyway, it's ok to keep pushing past this for now, but I think ultimately I _might_ prefer the `Arc<SpanValue<'arena>>` order.

fathom/src/surface/distillation.rs
wezm: Thanks for noticing and highlighting it. I've reworked the commits to not populate distilled `Terms` with a `Span`. It cleans up a lot of `TODO`/`FIXME` questions I had and doesn't impact the diagnostics that can be generated.

fathom/src/surface/elaboration.rs
```diff
@@ -1900,118 +1990,284 @@ impl<'interner, 'arena, 'error> Context<'interner, 'arena, 'error> {
         let operand_types = Option::zip(lhs_type.match_prim_spine(), rhs_type.match_prim_spine());
 
         let (fun, output_type) = match (op, operand_types) {
```
brendanzab, line 1992: It’s a bit of a shame that the formatting went out the window on the operator matches… really wish `rustfmt` let you add annotations to expressions for stuff like this 🤔

Won't hold off the PR for this but might have a look myself at some stage.

brendanzab (approved):
Looks good I think, lets merge this! Any other stuff can be addressed at a later point as we come across it.

PR #382, opened 2022-08-12
Add array*_index function

wezm:
This is once thing needed to be able to parse simple glyphs. It allows the number of coordinates in the glyph to be determined by accessing the last end point index.


doc/reference.md
```diff
@@ -994,11 +994,27 @@ fixed-length array type.
...
+`none`.
+
 - `array8_find : fun (len : U8) -> fun (A : Type) -> (A -> Bool) -> Array8 len A -> Option A`
 - `array16_find : fun (len : U16) -> fun (A : Type) -> (A -> Bool) -> Array16 len A -> Option A`
 - `array32_find : fun (len : U32) -> fun (A : Type) -> (A -> Bool) -> Array32 len A -> Option A`
 - `array64_find : fun (len : U64) -> fun (A : Type) -> (A -> Bool) -> Array64 len A -> Option A`
 
+#### index
+
+`array*_index` returns the item at the supplied index in the array.
```
mikeday, line 1011: Perhaps this needs to say what happens when the index is out of range?

fathom/src/core/semantics.rs
```diff
@@ -598,6 +598,22 @@ fn prim_step(prim: Prim) -> Option<PrimStep> {
...
+            step!(context, [_, _, index, array] => match array.as_ref() {
+                Value::ArrayLit(elems) => {
+                    let index = match (index).as_ref() {
+                        Value::ConstLit(Const::U8(index, _)) => Some(usize::from(*index)),
+                        Value::ConstLit(Const::U16(index, _)) => Some(usize::from(*index)),
+                        Value::ConstLit(Const::U32(index, _)) => usize::try_from(*index).ok(),
+                        Value::ConstLit(Const::U64(index, _)) => usize::try_from(*index).ok(),
+                        _ => return None,
+                    }?;
+                    elems[index].clone()
```
brendanzab, line 611: To be consistent with the other fallible ops I think we could change this to `elems.get(index).cloned()` - there's tradeoffs here but I think it’s probably better to avoid crashes in the evaluator that aren’t as a result of bugs in Fathom. 🤔

brendanzab (approved):
Looks good! 👍

PR #384, opened 2022-08-18
CI improvements and cleanups to flake.nix

brendanzab:
The PR reduces the number of Rust toolchains we test against on CI. This was a holdover from my habits during pre-1.0 Rust where we were encouraged to test _all the versions_. Hopefully Github’s error de-duplication will improve in the future (see https://github.com/actions/runner/issues/504).

I also took the time to clean up the `flake.nix` file. It skips the packaging for now and just loads a Rust development environment. This is used on CI to load Rust instead of [actions-rs](https://github.com/actions-rs/) which appears to be dormant. This might be controversial, but it seemed decently clean and decreases the amount of setup stuck in CI land. We can always switch to another approach in the future though, as Fathom’s use case is pretty simple.

I also added the minimum supported rust version to `fathom/Cargo.toml`. This is used to drive the MSRV used on CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,94 +10,63 @@ on:
       - main
 
 jobs:
-  spec:
-    runs-on: ubuntu-20.04
-    name: Check Makam Specification
-    steps:
-      - uses: actions/checkout@v2
-      - name: Install specification dependencies
-        run: yarn install
-        working-directory: ./experiments/makam-spec
-      - name: Check specification
-        run: yarn test
-        working-directory: ./experiments/makam-spec
-
-  check:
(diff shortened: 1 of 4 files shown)


wezm (approved):
Looks good (I'm going to assume the `flake.nix` is correct). Nice that the msrv from Cargo.toml is automatically tested without needing to update the CI config.

brendanzab:
@wezm Yeah, remembering to keep that in sync was always a pain! And it’s a bit finicky to pass to the build matrix.

Just pushed an update to do `cargo check`, `cargo build`, and `cargo test` all in one job to further pare things back.

Also happy to do a MSRV bump if you are finding the current version too restrictive?

PR #385, opened 2022-08-22
Clean up some formatting in match expressions

brendanzab:
Sometimes rustfmt gets a little excited with formatting, which has a negative impact on the readability of the code. To get around this we can either adjust the code to suit, or just use `#[rustfmt::skip]` to disable it in these cases.


wezm (approved):
Definitely a lot nicer

PR #386, opened 2022-08-23
Cleanups to arrow distillation

brendanzab:
Some more cleanups to arrow distillation

### DIFF
--- a/fathom/src/surface/distillation.rs
+++ b/fathom/src/surface/distillation.rs
@@ -370,7 +370,7 @@ impl<'interner, 'arena, 'env> Context<'interner, 'arena, 'env> {
             }
             core::Term::Universe(_span) => Term::Universe(()),
             core::Term::FunType(_span, _, input_type, output_type)
-                if !output_type.contains_free(LocalVar::last()) =>
+                if !output_type.binds_rigid_var(LocalVar::last()) =>
             {
                 let input_type = self.check(input_type);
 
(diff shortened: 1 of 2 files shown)


fathom/src/surface/distillation.rs
```diff
@@ -369,34 +369,33 @@ impl<'interner, 'arena, 'env> Context<'interner, 'arena, 'env> {
...
+                    self.pop_rigid();
+
+                    // Use an arrow type if the function type does not bind the
+                    // input name in the output type.
+                    Term::Arrow(
+                        (),
+                        self.scope.to_scope(input_type),
+                        self.scope.to_scope(output_type),
+                    )
+                }
```
wezm, line 398: Up to you but I think I would have kept this how it was since they share little code in the body and it avoids a level of indentation.

wezm (approved):

fathom/src/surface/distillation.rs
brendanzab: Hmm, ok, might invert the condition at least. I think my reasoning was that it keeps all the cases related to function types together - seeing as Rust is pretty verbose at times (especially given how it is used in Fathom) it can be hard to see that the patterns are related.

brendanzab: Ended up just adding a comment!

wezm: Nice. One of the talks I saw at the Brisbane Rust meetup a while back was about tips for making code more understandable to newcomers/your future self. One of the things the speaker said they liked to do was always comment the branches of a `match` like this. I thought it was a good idea.

PR #387, opened 2022-08-23
Rename environment structs in semantics

brendanzab:
This has been bugging me for a while, not sure what you think.

It’s probably a little less confusing to call these ‘environments’ as they are more related to evaluation, although to be honest the difference between ‘contexts’ and ‘environments’ is sometimes a tad… context dependent!

The rename reduces verbosity slightly in some cases which is a nice win for readability as well, I think.

### DIFF
--- a/fathom/src/core.rs
+++ b/fathom/src/core.rs
@@ -116,8 +116,8 @@ pub enum Term<'arena> {
     /// [entry information]: EntryInfo
     /// [function literals]: Term::FunLit
     /// [function applications]: Term::FunApp
-    /// [evaluation]: semantics::EvalContext::eval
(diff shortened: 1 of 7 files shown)


wezm (commented):
I don't have strong feelings one way or another. I think I've encountered "context" in some of the academic material I've (tried to) read. I've also encountered environment in the ECMAScript spec.

brendanzab:
Yeah there’s no single right answer on this. I think the reduced verbosity is nice enough, so let’s go with this and see how we feel later?

wezm:
> Yeah there’s no single right answer on this. I think the reduced verbosity is nice enough, so let’s go with this and see how we feel later?

Sure 👍

PR #388, opened 2022-08-23
Fix clippy lints

brendanzab:
Thought it was about time I fixed these clippy lints.

Seemed to uncover a bunch of reasonable issues even the there are some lints I’m a bit wary of.

### DIFF
--- a/fathom/src/core/semantics.rs
+++ b/fathom/src/core/semantics.rs
@@ -74,7 +74,7 @@ impl<'arena> Value<'arena> {
 
     pub fn match_prim_spine(&self) -> Option<(Prim, &[Elim<'arena>])> {
         match self {
-            Value::Stuck(Head::Prim(prim), spine) => Some((*prim, &spine)),
+            Value::Stuck(Head::Prim(prim), spine) => Some((*prim, spine)),
             _ => None,
         }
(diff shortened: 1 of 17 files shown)


fathom/src/core/semantics.rs
```diff
@@ -173,6 +174,7 @@ impl<'arena> Telescope<'arena> {
     }
 
     /// The number of terms in the telescope.
+    #[allow(clippy::len_without_is_empty)]
```
wezm, line 177: Maybe just turn this one off? Can't ever seeing it adding much value.

fathom/src/core/semantics.rs
```diff
@@ -164,6 +164,7 @@ impl<'arena> Telescope<'arena> {
         }
     }
 
+    #[allow(clippy::bool_assert_comparison)]
```
wezm, line 167: Maybe turn this one off too. Seems low-value.

wezm (approved):
Overall good improvements. I definitely agree there's some lints of dubious value that we might choose to disable.

brendanzab:
Think I’m going to enable clippy a separate PR. Some shenanigans is going on! 🤔

PR #390, opened 2022-08-23
Always return a function from prim_step

wezm:
I committed your idea from yesterday @brendanzab as I think it's an improvement.


brendanzab:
Awesome!

PR #391, opened 2022-08-25
Add some sketches of Fathom’s core in Idris 2

brendanzab:
Just a draft for now, but thought I’d publish if anyone was interested


experiments/idris/src/Fathom/Base.idr
```diff
@@ -0,0 +1,107 @@
...
+refine value {prf} = MkRefine { value, prf }
+
+
+||| Singleton types
+|||
+||| Inspired by [this type](https://agda.readthedocs.io/en/v2.5.4.1/language/with-abstraction.html#the-inspect-idiom)
+||| from the Agda docs.
+public export
+data Sing : {0 A : Type} -> (x : A) -> Type where
+  MkSing : {0 A : Type} -> {0 x : A} -> (y : A) -> {auto prf : x = y} -> Sing x
```
brendanzab, line 36: As strongly as I dislike Agda’s generalisation of lower case names in type signatures, it’s kind of forcing my hand due to the lack of sugar for dependent function types, or the ability to drop annotations for implicit arguments. This would be more cleanly defined as:

```idris
data Sing : (x : a) -> Type where
  MkSing : (y : a) -> {auto prf : x = y} -> Sing x
```

PR #393, opened 2022-09-02
Add bin ops to the reference

wezm:
As mentioned in https://github.com/yeslogic/fathom/pull/378#pullrequestreview-1035200218 the binary operators weren't mentioned in the reference, now they are :)


brendanzab:
Oh this looks great, thanks!

PR #395, opened 2022-09-08
Avoid needing to locally disable rustfmt

brendanzab:


brendanzab:
Merged seeing as it was relatively minor

PR #396, opened 2022-09-08
Span related cleanups

brendanzab:
Addressed some stuff I came across after the addition of spans. I’ve computed spans in some places in the elaborated terms. I also added a by-value conversion on byte ranges to make things a little cleaner in places, and cached some commonly used values to increase sharing (see the commit comment for more details).


wezm (approved):
Thanks for cleaning this up. I had intended the FIXMEs to be resolved before merging the original PR but got carried away and merged it with the still outstanding.

PR #401, opened 2022-09-27
Overhaul environment nomenclature

brendanzab:
As far as I can tell the flex/rigid naming was introduced in Huet (1975), “A Unification Algorithm for Typed lambda-Calculus”, but it ends up reading weirdly in the context of Fathom. Hoping that this makes the internals a little less bewildering/confusing.


wezm (approved):
Looks good.

We discussed the local/meta variable renaming. Not sure I completely understand the index and level names..?

brendanzab:
Yeah I really hate the naming of de Bruijn indices/levels, but I'm not sure of a better alternative right now. Using local/global was an attempt to come up with something nicer but I found it harder to remember, and conflicts with the  naming of the local environment.

Kmeakin:
> Yeah I really hate the naming of de Bruijn indices/levels (they don't really tell you all that much), but I'm not sure of a better alternative right now. Using local/global was an attempt to come up with something nicer but I found it harder to remember, and conflicts with the naming of the local environment. Happy to revisit at another time though.

How about "relative"/"absolute" in place of the "index"/"level" nomenclature?

brendanzab:
Ooooh interesting! That's neat. Reminds me that I've considered `VarOffset`/`VarId` in the past. 🤔

PR #402, opened 2022-10-06
Unfold metas

brendanzab:
This implements metavariable unfolding (a.k.a. _zonking_), which unfolds unification solutions after elaborating a module. This is useful for decoupling the elaborated modules and terms from the metacontext, which is important when rendering error messages and performing compilation down-the-line. I used an approach similar to [the one in Andras Kovacs’ staged repository](https://github.com/AndrasKovacs/staged/blob/20ccc4fe960e0ab24499bb639fef6fc80182dd0e/demo/Evaluation.hs#L128-L170). 

Note that some metavariables unfold to [large terms](https://github.com/yeslogic/fathom/pull/402/files#diff-b643aa35571f006ec8ba70a5dd46cd1f9dc5f4d0d9495caea3be4bbcbf550d44L1068-R1094). This is something that might be improved with glued evaluation. Something simple like [the one demoed in the elaboration-zoo](https://github.com/AndrasKovacs/elaboration-zoo/blob/e7c46a14270cff26b201dda14f1d80fa5d7d16cf/GluedEval.hs) could work for us here.

## References

- [The GHC Commentary: Type Variables and Zonking](https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/compiler/type-checker#types-variables-and-zonking)
- [What does the GHC source mean by "zonk"?](https://stackoverflow.com/questions/31889048/what-does-the-ghc-source-mean-by-zonk)
- [Email by Simon Peyton-Jones](https://mail.haskell.org/pipermail/glasgow-haskell-users/2013-August/024209.html)


brendanzab:
I think this is ready! There's more I'd like to do on this, but I think it’s an improvement.

wezm (approved):

PR #404, opened 2022-10-08
Add some assertions that all `Term`s produced by elaboration and quotatation are well-scoped

Kmeakin:
Could also be done with [`contracts`](https://docs.rs/contracts/latest/contracts/) if we don't mind adding an extra dependency.


Kmeakin:
Abandoning

PR #406, opened 2022-10-29
Add if-then-else sugar for match expressions

Kmeakin:
Adds if-then-else surface syntax that is elaborated to a `match` expression over `Bool`:

```
if c then t else f
==>
match c {
    true => t,
    false => f,
}
```

`match` expressions over `Bool` are also distilled to if-then-else expressions:
```
match c {
    true => t,
    false => f,
}
==>
if c then t else f
```

I wasn't sure what priority if-then-else should be parsed/printed with, so I gave it the same priority as `let` expressions for now, but it can be changed.


wezm:
This looks good to me but I'll leave it to Brendan to do final review when he's back from public holiday tomorrow.

fathom/src/surface/pretty.rs
```diff
@@ -424,6 +438,151 @@ impl<'interner, 'arena> Context<'interner, 'arena> {
...
+        cond_expr: &Term<'_, Range>,
+        then_expr: &Term<'_, Range>,
+        else_expr: &Term<'_, Range>,
+    ) -> DocBuilder<'arena, Self> {
+        let cond = docs![
+            self,
+            "if",
+            self.space(),
+            self.term_prec(Prec::Fun, cond_expr),
+        ];
```
brendanzab, line 454: Generally I’ve been using `self.concat([ ... ])` for this. Is it ok if we switch to that for consistency?

brendanzab:
Looks good! Just a minor thing on the pretty printing.

brendanzab:
Thanks!

PR #407, opened 2022-10-30
Add tuple syntax sugar for records

Kmeakin:
Adds tuple syntax that is elaborated to record literals/record types/record formats:

`(e0, e1, ...)` synthesises to `{_0 = e0, _1 = e1, ...}`
`(e0, e1, ...)` checks against `Type` to `{_0: e0, _1: e1, ...}`
`(e0, e1, ...)` checks against `Format` to `{_0 <- e0, _1 <- e1, ...}`
`(e0, e1, ...)` checks against `{l0: T0, l1: T1, ...}` to `{l0 = e0, l1 = e1, ...}`

This also makes the `surface::Term::UnitLiteral` production unecessary. `{}` is now parsed as an empty tuple (meaning that `{}` and `()` are indistinguishable, as in Rust).


fathom/src/surface/elaboration.rs
```diff
@@ -965,17 +965,73 @@ impl<'interner, 'arena> Context<'interner, 'arena> {
...
+                    && matches!(
+                        expected_type.match_prim_spine(),
+                        Some((Prim::FormatType, [])),
+                    ) =>
             {
                 core::Term::FormatRecord(range.into(), &[], &[])
             }
+            (Term::Tuple(range, elem_exprs), Value::Universe) => {
+                let labels = (0..elem_exprs.len())
+                    .map(|idx| self.interner.borrow_mut().get_or_intern(format!("_{idx}")));
```
wezm, line 979: Looks CI failed because our minimum supported Rust version doesn't support the format string shorthand. If you could tweak that to use the more verbose form it should fix it.

fathom/src/surface/elaboration.rs
```diff
@@ -965,17 +965,73 @@ impl<'interner, 'arena> Context<'interner, 'arena> {
...
+                    {
+                        expr_labels.push((elem_expr.range(), *label));
+                        elem_exprs.next();
+                    }
+
+                    // use numeric labels for excess elems
+                    for (idx, elem_expr) in elem_exprs {
+                        expr_labels.push((
+                            elem_expr.range(),
+                            self.interner.borrow_mut().get_or_intern(format!("_{idx}")),
```
wezm, line 1009: and here

wezm (changes requested):

wezm (commented):
Thanks for the update, looks good

fathom/src/surface/grammar.lalrpop
```diff
@@ -199,7 +199,7 @@ ProjTerm: Term<'arena, ByteRange> = {
 };
 
 AtomicTerm: Term<'arena, ByteRange> = {
-    "(" <term: Term> ")" => term,
+    <start: @L> "(" <terms: Seq<Term, ",">> ")" <end: @R> => if terms.len() == 1 { terms[0].clone() } else { Term::Tuple(ByteRange::new(file_id, start, end), terms) },
```
brendanzab, line 202: This means that `"(" <Term> "," ")"` gets parsed as a parenthesized term. Unsure if this is what we want. I think maybe we should have separate rules for: `"(" ")"`, `"(" <Term> ")"`, `"(" <Term> "," ")"` and `"(" <Term> "," <Seq<Term, ",">> ")"`?

Kmeakin: Sure!

fathom/src/driver.rs
```diff
@@ -365,6 +365,7 @@ impl<'surface, 'core> Driver<'surface, 'core> {
                 context.text("="),
                 context.space(),
                 context.sequence(
+                    true,
```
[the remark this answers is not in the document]
brendanzab: Oh I see, it adds a space to the inside of the delimiters. 🤔

brendanzab:
Might also be worth trying to rebase this to some smaller logical commits once you are done, but looks close to being mergeable.

Would also be a good idea to mention tuples in the [language reference](https://github.com/yeslogic/fathom/blob/main/doc/reference.md) too. Just remembered!

brendanzab:
Eventually we might consider adding tuple format sugar too: ie. `(e0, e1, ...)` could check against `Format`, elaborating to a record format.

fathom/src/env.rs
```diff
@@ -42,7 +42,7 @@ type RawVar = u16;
 /// [de Bruijn index]: https://en.wikipedia.org/wiki/De_Bruijn_index
 /// [alpha-equivalence]: https://ncatlab.org/nlab/show/alpha-equivalence
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-pub struct Index(RawVar);
+pub struct Index(pub RawVar);
```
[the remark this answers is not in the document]
Kmeakin: Must be a leftover from experimenting with implementations of `is_record_type` which I forgot to clean up.

fathom/src/surface/grammar.lalrpop
```diff
@@ -299,8 +306,14 @@ Seq<Elem, Sep>: &'arena [Elem] = {
...
     <first: Elem> <elems: (Sep <Elem>)*> Sep? => {
         scope.to_scope_from_iter(std::iter::once(first).chain(elems))
     },
 };
+
+Seq2<Elem, Sep>: &'arena [Elem] = {
+    <first: Elem> Sep <second: Elem> <elems: (Sep <Elem>)*> Sep? => {
+        scope.to_scope_from_iter(std::iter::once(first).chain(std::iter::once(second)).chain(elems))
+    },
+};
```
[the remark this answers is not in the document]
Kmeakin: Ah, I see you have this set in `.editconfig`. Might be worth adding editorconfig as a recommended extension for VSCode users!

brendanzab:
Thanks so much!

PR #409, opened 2022-11-02
Record and tuple patterns

Kmeakin:
Adds record and tuple patterns. When used in "irrefutable" positions (ie `let` expressions, fun literals and fun types) they elaborate to chains of simpler `let` expressions.

```fathom
let {x = a, y = b} = r;
z
```
elaborates to
```fathom
let a = r.x;
let b = r.y;
z
```

Using record/tuple patterns in `match` expressions is much more complex, so I will leave that for a follow-up PR

### DIFF
--- a/fathom/src/core.rs
+++ b/fathom/src/core.rs
@@ -1,6 +1,8 @@
 //! Core language.
 
-use crate::env::{Index, Level};
+use scoped_arena::Scope;
+
+use crate::env::{EnvLen, Index, Level};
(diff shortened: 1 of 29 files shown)


Kmeakin:
Abandoning for now

brendanzab:
Sorry for my slowness - yeah I think I’d prefer introducing weakening in the core language for something like this, but it’s a valuable attempt!

PR #410, opened 2022-11-05
Fix issue #408

### LINKED ISSUE #408: Inconsistency between synthesised function literals and checked function literals
The behaviour of function literals changes depending on whether they are synthesised or checked during elaboration.

eg:
`fun (A : Type) (B : Type) (p : {x : A, y : B}) => p.x`
elaborates to `fun A B p => p.a : fun (A : Type) (B : Type) -> { a : A, b : B } -> A`

but
`(fun (A : Type) (B : Type) (p : {x : A, y : B}) => p.x) : _`
elaborates to `fun A B p => p.a : fun (A : Type) -> Type -> { a : A, b : a } -> A`

Kmeakin:
The cause of the bug was in the implementation of renaming record telescopes: we were pushing variables onto the local environment while traversing the telescope, rather than onto the partial renaming (cf `rename_closure`)

Fixes #408

### DIFF
--- a/fathom/src/surface/elaboration/unification.rs
+++ b/fathom/src/surface/elaboration/unification.rs
@@ -643,7 +643,7 @@ impl<'arena, 'env> Context<'arena, 'env> {
         meta_var: Level,
         telescope: &Telescope<'arena>,
     ) -> Result<&'arena [Term<'arena>], RenameError> {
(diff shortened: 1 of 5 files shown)


brendanzab:
Thanks!

PR #414, opened 2022-11-10
Unify constant matches

brendanzab:
Somewhere along the line we forgot to add a case for this.

### DIFF
--- a/fathom/src/core.rs
+++ b/fathom/src/core.rs
@@ -175,9 +175,8 @@ pub enum Term<'arena> {
 
     /// Constant literals.
     ConstLit(Span, Const),
-    /// Match on a constant.
-    ///
-    /// (head_expr, pattern_branches, default_expr)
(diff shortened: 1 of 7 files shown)


Kmeakin:
Could we have a more strongly normalizing representation for branches? At the moment I believe these two functions won't unify:

```fathom
let not1 = fun (b : Bool) => match b {
    true => false,
    _ => true,
}

let not2 = fun (b : Bool) => match b {
    false => true,
    _ => false,
}
```

brendanzab:
Hmm, not sure! At the moment the branches are elaborated into a sorted list of cases, but yeah you are right there’s cases where that falls down due to the default branch

brendanzab:
Ahh I see what you mean. Yeah if expressions are currently desugared out of order (fixed that).

wezm (approved):
Looks good 👍

PR #418, opened 2022-11-14
Some tweaks to tuple distillation

brendanzab:
Some tweaks to the distillation of tuples:

- parses the field labels as opposed to pretty printing then checking for equality
- only checks for type dependencies if the fields of a record look like a tuple

### DIFF
--- a/fathom/src/core.rs
+++ b/fathom/src/core.rs
@@ -1,8 +1,7 @@
 //! Core language.
 
 use crate::env::{Index, Level};
-use crate::source::Span;
-use crate::StringId;
+use crate::source::{Span, StringId};
 
 pub mod binary;
 pub mod prim;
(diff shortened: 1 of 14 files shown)


Kmeakin:
I have another idea for avoiding allocations and parsing:

```rust
pub struct Interner {
    strings: StringInterner,
    tuple_labels: Vec<StringId>,
}

impl Interner {
    pub fn get_tuple_label(&self, index: usize) -> StringId { self.tuple_labels[index] }
    pub fn ensure_tuple_label(&mut self, max_index: usize) {...}
}
```

This would allow us to avoid parsing strings when distilling tuples, and to avoid allocating and formatting strings when elaborating tuples.

Kmeakin:
Alternatively, we could parse field labels into integers during parsing:

```rust
pub enum FieldLabel {
    Ident(StringId),
    Tuple(u16)
}
```

PR #419, opened 2022-11-25
Preserve name of default branch in `core::Term::ConstMatch`

Kmeakin:
At the moment, we throw away the name of the binder when elaborating pattern matches to `ConstMatch` with a default branch

eg
```fathom
match x {
    0 => 1,
    y => y,
}
```
elaborates to 
```fathom
match x {
    0 => 1,
    _ => _,
}
```
This patch preserves the name of the binding, making the elaborated core more readable


brendanzab:
This is great, thanks!

PR #422, opened 2022-12-02
Fix issue #421

### LINKED ISSUE #421: Inconsistency between tuple types and record types
The behaviour of record types changes depending on whether their labels are tuple-like or not

eg:
```fathom
let fst = fun (A: Type) (B: Type) (p: {x: A, y: B}) => p.x;
```
elaborates to
```fathom
let fst : fun (A : Type) (B : Type) -> { x : A, y : B } -> A = fun A B p => p.x;
```

but
```fathom
let fst = fun (A: Type) (B: Type) (p: {_0: A, _1: B}) => p._0;
```
elaborates to
```fathom
let fst : fun (A : Type) (B : Type) -> (A, A) -> A = fun A B p => p._0
```
note the pair is incorrectly displayed as `(A, A)` instead of `(A, B)`

Kmeakin:
- The issue was in distillation: we didn't push labels while distilling dependent tuples.

- Fixed a bug in the caching of tuple labels: `get_tuple_label` would not grow the `tuple_labels` vector, so `tuple_labels` would always be empty! This didn't affect correctness, but it did mean there was no performance improvement. 

- Added `get_tuple_labels` for getting a whole slice of tuple labels at once, and used it to optimize the implementation of `is_tuple_labels` 

Fixes #421

### DIFF
--- a/fathom/src/source.rs
+++ b/fathom/src/source.rs
@@ -1,6 +1,6 @@
 //! Types related to source files.
 
-use std::ops::{Deref, DerefMut};
+use std::ops::{Deref, DerefMut, Range};
 
 // Interned strings.
 pub type StringId = string_interner::symbol::SymbolU16;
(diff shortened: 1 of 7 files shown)


brendanzab:
Oh, woops! Did I introduce this?

fathom/src/source.rs
```diff
@@ -53,7 +72,7 @@ impl StringInterner {
 
     /// Returns true if `labels` is a sequence of tuple labels: `_0`, `_1`, ...
     pub fn is_tuple_labels(&mut self, labels: &[StringId]) -> bool {
-        (labels.iter().enumerate()).all(|(index, label)| self.is_tuple_label(index, *label))
+        labels == self.get_tuple_labels(0..labels.len())
     }
```
brendanzab, line 76: Cool idea!

brendanzab:
Looks good, thanks for catching this!

PR #423, opened 2022-12-02
Add pretty printer for core language

wezm:
To make it easier to view things in the core language I created a pretty printer for it, heavily derived from `surface::pretty`. My intention is that it's mostly a debugging tool/a nicer version of the `Debug` output. Output looks like this:

```
def empty : FormatType = {};

def kern : FormatType = (let kern0 : FormatType = {
    version <- { version <- FormatU16Be | (U16Eq LocalVar(Index(0))) (U16(0, Decimal) : U16Type) },
    num_tables <- FormatU16Be,
    subtables <- (FormatArray16 LocalVar(Index(0))) ItemVar(Level(2)),
};
let kern1 : FormatType = {
    version <- { version <- FormatU32Be | (U32Eq LocalVar(Index(0))) (U32(65536, Hexadecimal) :
    U32Type) },
    num_tables <- FormatU32Be,
    subtables <- (FormatArray32 LocalVar(Index(0))) ItemVar(Level(3)),
};
{
    kern <- overlap {
        version <- FormatU16Be,
        table <- match LocalVar(Index(0)) {
            U16(0, Decimal) => LocalVar(Index(2)),
            U16(1, Decimal) => LocalVar(Index(1)),
            _ => ItemVar(Level(4)),
        },
    },
});

def main : FormatType = ItemVar(Level(5));
```


fathom/src/core/pretty.rs
```diff
@@ -0,0 +1,470 @@
...
+    fn term_prec(&'arena self, prec: Prec, term: &Term<'arena>) -> DocBuilder<'arena, Self> {
+        // FIXME: indentation and grouping
+
+        match term {
+            Term::ItemVar(_, level) => self.text(format!("ItemVar({:?})", level)),
+            Term::LocalVar(_, index) => self.text(format!("LocalVar({:?})", index)),
+            Term::MetaVar(_, index) => self.text(format!("MetaVar({:?})", index)),
+            Term::InsertedMeta(_, level, info) => {
+                self.text(format!("InsertedMeta({:?}, {:?})", level, info))
+            }
```
brendanzab, line 135: Not a big issue, but perhaps we could do something a bit more compact for variables? Like `item#n`, `local#n`... (not sure of the naming or the syntax there). Another thought is that we could also add an environment of binder names for items, locals, and metas to make things more readable. Eg. `name#n` (still showing the level/index for debugging purposes). Not a massive issue though, just a thought!

fathom/src/core/pretty.rs
```diff
@@ -0,0 +1,470 @@
...
+        self.scope.to_scope(f)
+    }
+
+    fn alloc_width_fn(
+        &'arena self,
+        f: impl 'arena + Fn(isize) -> Self::Doc,
+    ) -> <Self::Doc as DocPtr<'arena, A>>::WidthFn {
+        self.scope.to_scope(f)
+    }
+}
```
brendanzab, line 470: I wonder if we could skip this optimisation seeing as this is just for debugging. No drama though either way.

brendanzab (approved):
Looks good feel free to merge as you wish. No need to address any of my comments unless you feel you want to.

fathom/src/core/pretty.rs
Kmeakin: If the core pretty printer does include names, I would like a command line option to disable human names and only print indices: in my experience scoping/naming errors are very common. A pretty printer outputting incorrect names would be worse than useless when debugging!

brendanzab: Yeah that makes sense!

wezm:
I tweaked the formatting of vars, switched to using `RcDoc` instead of custom `DocAllocator` impl, and hooked it up to the CLI.

brendanzab (approved):

PR #427, opened 2022-12-03
Reserve extra capacity upfront in more places

Kmeakin:


wezm (approved):
I thought I do some testing with [dhat-rs](https://docs.rs/dhat/latest/dhat/) to see what sort of difference it makes.

I profiled `cargo r --release --features dhat-heap -- data --module formats/opentype.fathom Verdana.ttf`

**Before:**

```
dhat: Total:     60,143,379 bytes in 281,457 blocks
dhat: At t-gmax: 36,074,654 bytes in 128,913 blocks
dhat: At t-end:  1,374 bytes in 3 blocks
```

**Note:** `dhat` calls an individual allocation a "block"

**After:**

```
dhat: Total:     60,097,141 bytes in 281,433 blocks
dhat: At t-gmax: 36,064,240 bytes in 128,913 blocks
dhat: At t-end:  1,374 bytes in 3 blocks
```

Some logging at the `reserve` call site suggests that a lot of the time the vector already had capacity for the new items being added by having a capacity of 16 or 32 so the reserve ended up being a no-op.

Anyway seems like a reasonable change nonetheless and nice to have.

PR #429, opened 2022-12-11
Apply clippy suggestions

Kmeakin:
Updating the minimum rust version to `1.62.0` triggered some more clippy lints


wezm:
Thanks!

PR #430, opened 2022-12-12
Record/tuple patterns and new pattern match compiler and coverage checker

Kmeakin:
* Adds record and tuple patterns. 
* This required rewriting the pattern match compiler and coverage checker, following the algorithms outlined in [Compiling pattern matching to good decision trees](https://dl.acm.org/doi/10.1145/1411304.1411311) and [Warnings for pattern matching](http://moscova.inria.fr/~maranget/papers/warn/index.html), respectively
* The new algorithms can be extended in the future to support more pattern matching features, such as or-patterns and pattern guards.

# Future work
* I have not done any optimization: there are a lot of allocations that could probably be eliminated.
* The coverage checker simply reports that a match is in-exhaustive, but does not explain which patterns are missing. A more sophisticated algorithm for calculating missing patterns in described in the second half of "Warnings for pattern matching"
* The match compiler produces ugly code: it inserts extra let bindings in the body of default branches, instead of binding the default branch to a variable:
```fathom
match x {
    0 => 0,
    y => y + 1,
}
```
will become
```fathom
match x {
    0 => 0,
    _ => let y = x; y + 1,
}
```


fathom/src/surface/elaboration/patterns/compile.rs
```diff
@@ -0,0 +1,114 @@
+//! # Bibliography
+//! - [Compiling pattern matching to good decision trees](https://dl.acm.org/doi/10.1145/1411304.1411311)
+//!     - https://github.com/Lapz/Compiling-Pattern-Matching-to-Good-Decision-Trees
+//!     - https://github.com/SomewhatML/match-compile
+//! - [How to compile pattern matching](https://julesjacobs.com/notes/patternmatching/patternmatching.pdf)
+//!     - https://gitlab.com/yorickpeterse/pattern-matching-in-rust/-/tree/main/jacobs2021
```
wezm, line 6: Bare links need to be wrapped in `< >`.

```
warning: this URL is not a hyperlink
 --> fathom/src/surface/elaboration/patterns/compile.rs:3:11
  |
3 | //!     - https://github.com/Lapz/Compiling-Pattern-Matching-to-Good-Decision-Trees
  |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use an automatic link instead: `<https://github.com/Lapz/Compiling-Pattern-Matching-to-Good-Decision-Trees>`
  |
  = note: `#[warn(rustdoc::bare_urls)]` on by default
  = note: bare URLs are not automatically turned into clickable links
```

wezm (commented):
I'll leave a full review of this to when Brendan returns from leave. Just noting one small issue with links in docs.

brendanzab (commented):
Still need to look deeper into this to try to understand the algorithm… there’s a bit to take in but it’s looking neat!

fathom/src/core.rs
```diff
@@ -255,6 +261,207 @@ impl<'arena> Term<'arena> {
             }
         }
     }
+
+    // TODO: Add a new `Weaken` variant to `core::Term` instead of eagerly traversing the term?
```
brendanzab, line 265: For the benefit of others, I believe Andras Kovacs’ staged language [has something similar](https://github.com/AndrasKovacs/staged/blob/9e381eb162f44912d70fb843c4ca6567b0d1683a/demo/Syntax.hs#L52)?

brendanzab:
We could maybe pull out the rust upgrade into a separate PR if it helps clean things up.

fathom/src/surface/elaboration.rs
```diff
@@ -1955,217 +1947,60 @@ impl<'interner, 'arena> Context<'interner, 'arena> {
...
-        // Report if we can still reach this point
-        if is_reachable {
-            // TODO: this should be admitted if the scrutinee type is uninhabited
-            self.push_message(Message::NonExhaustiveMatchExpr {
-                match_expr_range: match_info.range,
-                scrutinee_expr_range: match_info.scrutinee.range,
-            });
-        }
-        core::Term::Prim(match_info.range.into(), Prim::ReportedError)
+        debug_assert_eq!(matrix.num_rows(), bodies.len());
```
brendanzab, line 2001: I’d be cool if we added [`contracts`](https://lib.rs/crates/contracts) to attach preconditions like this!

fathom/src/surface/elaboration/patterns/compile.rs
```diff
@@ -0,0 +1,210 @@
+//! # Bibliography
+//! - [Compiling pattern matching to good decision trees](https://dl.acm.org/doi/10.1145/1411304.1411311)
+//!     - <https://github.com/Lapz/Compiling-Pattern-Matching-to-Good-Decision-Trees>
+//!     - <https://github.com/SomewhatML/match-compile>
+//!     - <https://alan-j-hu.github.io/writing/pattern-matching.html>
+//! - [How to compile pattern matching](https://julesjacobs.com/notes/patternmatching/patternmatching.pdf)
+//!     - <https://gitlab.com/yorickpeterse/pattern-matching-in-rust/-/tree/main/jacobs2021>
```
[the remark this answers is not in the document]
Kmeakin: I think dependent pattern matching should leave the compilation to decision trees unaffected, except maybe inserting coercions. We will need to modify checking/synthesizing of patterns though.

fathom/src/surface/elaboration.rs
Kmeakin: In my experience, `contracts` can clutter backtraces and may cause compile errors in functions with lifetime annotations.

fathom/src/surface/distillation.rs
```diff
@@ -799,15 +844,24 @@ impl<'interner, 'arena, 'env> Context<'interner, 'arena, 'env> {
...
 
+#[allow(clippy::type_complexity)]
 fn match_if_then_else<'arena>(
     branches: &'arena [(Const, core::Term<'arena>)],
     default_branch: Option<(Option<StringId>, &'arena core::Term<'arena>)>,
-) -> Option<(&'arena core::Term<'arena>, &'arena core::Term<'arena>)> {
+) -> Option<(
+    (Option<Option<StringId>>, &'arena core::Term<'arena>),
+    (Option<Option<StringId>>, &'arena core::Term<'arena>),
+)> {
```
brendanzab, line 854: I’m… not really understanding what’s going on with the `Option<Option<StringId>>`s for the then and else branches?

brendanzab:
Sorry for the slowness - taking a bit of time to get into this one

fathom/src/surface/distillation.rs
Kmeakin: In cases of the form
```fathom
match x {
    true => ...,
    _ => ...,
}
```
or
```fathom
match x {
    false => ...,
    _ => ...,
}
```

We have to push `_` onto `local_names`, or else we get confusing bugs in distillation output. Alternatively, we could only distill matches to if-then-else syntax if they are of the form
```fathom
match x {
    true => ...,
    false => ...,
}
```

tests/fail/elaboration/boolean-literal/not-supported.snap
```diff
@@ -12,4 +12,10 @@ error: boolean literal not supported for expected type
 7 │     true => 1,
   │     ^^^^
 
+warning: unreachable pattern
+  ┌─ tests/fail/elaboration/boolean-literal/not-supported.fathom:7:5
+  │
+7 │     true => 1,
+  │     ^^^^
+
```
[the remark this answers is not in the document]
Kmeakin: Can do

brendanzab:
I’ve been looking at this more in depth locally. Doing a bit of messing around to understand it a bit more and tweaking the implementation a bit. While it’s big, I think it would be better to squash it into a single commit as it kind of only works all together.

One nit-pick: could we use the actual the actual type names rather than `Self` in impls and type definitions etc? I prefer to only use those when necessary in `trait` definitions.

Kmeakin:
> One nit-pick: could we use the actual the actual type names rather than `Self` in impls and type definitions etc? I prefer to only use those when necessary in `trait` definitions.

I think the preferred Rust style is to use `Self` wherever possible: https://rust-lang.github.io/rust-clippy/master/#use_self. Personally I prefer `Self` because:
* Usually less typing: most type names are more than 4 letters
* One less piece of information you need to keep track off when writing methods/impls. Never do something yourself that the compiler can do for you!
* Type names may change frequently during development. `Self` means you don't have to update all the occurrences of the type name, and keeps diffs smaller

fathom/src/surface/elaboration/patterns.rs
```diff
@@ -0,0 +1,755 @@
...
+        ctx: &mut elaboration::Context<'_, 'arena>,
+        ctor: &Constructor,
+        scrut: &Scrutinee<'arena>,
+    ) -> Option<PatRow<'arena>> {
+        match self {
+            CheckedPattern::ReportedError(range)
+            | CheckedPattern::Placeholder(range)
+            | CheckedPattern::Binder(range, _) => {
+                let columns =
+                    vec![(CheckedPattern::Placeholder(*range), scrut.clone()); ctor.arity()];
```
brendanzab, line 655: Why is it ok to duplicate these without updating `scrut`?

Kmeakin: A name/placeholder/error pattern leaves the scrutinee unchanged: it does not project further into it. It is not strictly necessary to duplicate the column: wildcard patterns could be removed entirely from the matrix (indeed, that is how it is done in [How to compile pattern matching](https://julesjacobs.com/notes/patternmatching/patternmatching.pdf)). But the algorithm as presented in [Compiling pattern matching to good decision trees](https://dl.acm.org/doi/10.1145/1411304.1411311) relies on the matrix always being rectangular: each row must have the same number of columns.

brendanzab:
I’ve looked into merging this, trying to do some cleaning up (see: https://github.com/Kmeakin/fathom/compare/pattern-matching...brendanzab:fathom:pattern-matching-cleanups), but I don’t think it is ready yet. Some of the pattern matching compilation is self-contained and can be iterated on, but my big fear is how it impacts the handling of local bindings in the elaborator. The `freshen_scrutinee` and `insert_extra_let` methods are of particular concern, and I’m a little worried about taking on this burden going forward. Related to #488, it does make me think we might need a better approach to managing local bindings in the elaborator - eg. regarding adding/removing bindings, or with weakening.

PR #435, opened 2023-01-01
Replace all used occurances of `_` with fresh names

Kmeakin:
Generate fresh names for used `_` names during distillation.

### DIFF
--- a/fathom/src/surface/distillation.rs
+++ b/fathom/src/surface/distillation.rs
@@ -86,6 +86,42 @@ impl<'interner, 'arena, 'env> Context<'interner, 'arena, 'env> {
         }
     }
 
+    /// Generate a fresh name that does not appear in `self.local_names`
+    fn gen_fresh_name(&mut self) -> StringId {
+        fn to_str(x: u32) -> String {
+            let base = x / 26;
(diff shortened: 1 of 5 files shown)


fathom/src/surface/distillation.rs
```diff
@@ -86,6 +86,42 @@ impl<'interner, 'arena, 'env> Context<'interner, 'arena, 'env> {
...
+        fn to_str(x: u32) -> String {
+            let base = x / 26;
+            let letter = x % 26;
+            let letter = (letter as u8 + b'a') as char;
+            if base == 0 {
+                format!("{letter}")
+            } else {
+                format!("{letter}{base}")
+            }
+        }
```
brendanzab, line 100: What is the behavior of this when `x > 26`? Is it something like: `a`, …, `z`, `a0`, …, `z0`, `a1`, …, `z1`, …?

brendanzab:
This looks much nicer, and less confusing!

fathom/src/surface/distillation.rs
Kmeakin: Yes

PR #437, opened 2023-01-06
Use raw identifiers in LALRPOP grammar

Kmeakin:
I opened a PR against LALRPOP to lex raw identifiers properly: https://github.com/lalrpop/lalrpop/pull/698. If it is accepted, we can use raw identifiers in our `grammar.lalrpop`.

### DIFF
--- a/fathom/Cargo.toml
+++ b/fathom/Cargo.toml
@@ -32,7 +32,7 @@ string-interner = "0.14.0"
 termsize = "0.1.6"
 
 [build-dependencies]
-lalrpop = "0.19.5"
+lalrpop = { git = "https://github.com/kmeakin/lalrpop", branch = "raw-identifiers" }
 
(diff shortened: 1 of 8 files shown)


fathom/Cargo.toml
```diff
@@ -32,7 +32,7 @@ string-interner = "0.14.0"
 termsize = "0.1.6"
 
 [build-dependencies]
-lalrpop = "0.19.5"
+lalrpop = { git = "https://github.com/kmeakin/lalrpop", branch = "raw-identifiers" }
```
brendanzab, line 35: Will obviously need to be published if we want to eventually publish Fathom, but that’s a long way off, so I think it’s ok for now!

brendanzab (approved):
Sweet! This has been an annoyance for a while now!

PR #438, opened 2023-01-06
Lex raw identifiers

Kmeakin:
Allow keywords to be used as identifiers by prefixing them with `r#`, as in Rust (eg `r#let`)

### DIFF
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -117,6 +117,8 @@ Names start with a an ASCII letter (`a`..`z` or `A`..`Z`), and end with a
 sequence of ASCII letters (`a`..`z` or `A`..`Z`), numbers (`0`..`9`), or
 underscores (`_`).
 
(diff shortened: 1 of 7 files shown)


brendanzab (changes requested):
I think it’d be good to update [the reference](#https://github.com/yeslogic/fathom/blob/main/doc/reference.md#names) to mention raw idents, but looks good otherwise! Might be nice to bikeshed some of this syntax at some stage, but this is nice to have.

brendanzab (approved):
Seems like the main branch was accidentally merged into this branch, as opposed to rebasing on top of main? But otherwise looks good!

Kmeakin:
> Seems like the main branch was accidentally merged into this branch, as opposed to rebasing on top of main? But otherwise looks good!

That's what I get for using the github web UI to resolve conflicts :facepalm:

PR #440, opened 2023-01-08
Fix ordering of `expected` and `found` types in elaboration::convert

Kmeakin:
The ordering of the "expected" and "found" types when reporting a type mismatch was the wrong way around.

# Example
```fathom
false : U32
```
would produce 
```text
type mismatch, expected `Bool`, found `U32`
```

the correct order is
```text
type mismatch, expected `U32`, found `Bool`
```

### DIFF
--- a/fathom/src/surface/elaboration.rs
+++ b/fathom/src/surface/elaboration.rs
@@ -456,8 +456,9 @@ impl<'interner, 'arena> Context<'interner, 'arena> {
             });
         }
 
-        let filtered_fields = (fields.iter().enumerate())
-            .filter_map(move |(index, field)| (!duplicate_indices.contains(&index)).then_some(field));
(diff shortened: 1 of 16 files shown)


fathom/src/surface/elaboration.rs
```diff
@@ -570,8 +571,8 @@ impl<'interner, 'arena> Context<'interner, 'arena> {
         &mut self,
         surface_range: ByteRange, // TODO: could be removed if we never encounter empty spans in the core term
         expr: core::Term<'arena>,
-        type0: &ArcValue<'arena>,
-        type1: &ArcValue<'arena>,
+        from: &ArcValue<'arena>,
+        to: &ArcValue<'arena>,
```
brendanzab, line 575: Oh yes, I prefer this naming 👍

brendanzab (approved):

brendanzab:
Great improvement, thanks!

PR #441, opened 2023-01-08
Remove unecessary `push_item` from distillation

Kmeakin:
`item_env.names` is already populated by elaboration, so we don't need to push item names onto the environment while distilling.


brendanzab (approved):
Thanks!

PR #442, opened 2023-01-09
Move alphabetic name creation into string interner

brendanzab:


fathom/src/source.rs
```diff
@@ -76,6 +114,21 @@ impl StringInterner {
     }
 }
 
+fn alphabetic_name(index: usize) -> String {
+    let base = index / 26;
+    let letter = index % 26;
```
wezm, line 119: It's weird that Rust doesn't have `divmod` particularly since x86 `div` provides both. I suppose LLVM might work that out though.

Ok yes, looks like it can work it out: https://rust.godbolt.org/z/hj9rb4Txa

wezm (approved):
Looks good 👍

fathom/src/source.rs
brendanzab: Lol, I think I actually added divmod to the stdlib _years_ ago. Maybe we removed at some stage before 1.0?

wezm: Amazing! Sadly no match for `div_mod` in the docs now.

brendanzab: It was named `div_rem` at some point too! https://github.com/rust-lang/rust/pull/6147/commits/ee26c7c433dbb8d41a2b65dbc89eb84acfc2d311

brendanzab: I think it might live in num-rs now 🤔

wezm: Well this has been a fun history lesson

PR #443, opened 2023-01-09
Reduce size of `ByteRange` and `Span`

Kmeakin:
By using `u32` instead of `usize` for `BytePos`, and `NonZeroU32` instead of `usize` for `FileId`, we can significantly reduce the size of `ByteRange` and `Span`. This also reduces the size of types that carry `ByteRange` or `Span`:

# Size reductions
- `source::ByteRange`: 24 bytes to 12 bytes
- `source::Span`: 24 bytes to 12 byes
- `surface::Term<ByteRange>`: 80 bytes to 56 bytes
- `surface::Pattern<ByteRange>`: 32 bytes to 16 bytes
- `core::Term`: 72 bytes to 56 bytes
- `semantics::Value`: 80 bytes to 72 bytes
- `elaboration::CheckedPattern`: 40 bytes to 32 bytes

### DIFF
--- a/fathom/src/core.rs
+++ b/fathom/src/core.rs
@@ -713,4 +713,10 @@ mod tests {
         assert!(!std::mem::needs_drop::<Term<'_>>());
         assert!(!std::mem::needs_drop::<Term<'_>>());
     }
+
+    #[test]
+    #[cfg(target_pointer_width = "64")]
+    fn term_size() {
+        assert_eq!(std::mem::size_of::<Term>(), 56);
+    }
 }
--- a/fathom/src/driver.rs
+++ b/fathom/src/driver.rs
@@ -7,7 +7,8 @@ use std::path::Path;
 
 use crate::core;
 use crate::core::binary::{self, BufferError, ReadError};
-use crate::source::{ByteRange, FileId, Span, StringInterner};
+use crate::files::{FileId, Files};
+use crate::source::{ByteRange, Span, StringInterner};
 use crate::surface::{self, elaboration};
 use crate::BUG_REPORT_URL;
 
@@ -27,7 +28,7 @@ impl Status {
 }
 
 pub struct Driver<'surface, 'core> {
-    files: SimpleFiles<String, String>,
+    files: Files<String, String>,
     interner: RefCell<StringInterner>,
     surface_scope: scoped_arena::Scope<'surface>,
     core_scope: scoped_arena::Scope<'core>,
@@ -47,7 +48,7 @@ impl<'surface, 'core> Driver<'surface, 'core> {
             interner: RefCell::new(StringInterner::new()),
             surface_scope: scoped_arena::Scope::new(),
             core_scope: scoped_arena::Scope::new(),
-            files: SimpleFiles::new(),
+            files: Files::new(),
 
             allow_errors: false,
             seen_errors: RefCell::new(false),
--- a/fathom/src/source.rs
+++ b/fathom/src/source.rs
@@ -2,6 +2,8 @@
 
 use std::ops::{Deref, DerefMut, Range};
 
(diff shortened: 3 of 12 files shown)


fathom/src/core.rs
```diff
@@ -185,6 +185,11 @@ pub enum Term<'arena> {
     ),
 }
 
+#[test]
+fn term_size() {
+    assert_eq!(std::mem::size_of::<Term>(), 56);
+}
+
```
brendanzab, line 192: I think I tend to prefer to put unit tests at the end of a file, in a submodule - eg. https://github.com/yeslogic/fathom/blob/9cbfe12a1e102b06118f1299beae908ce0fc4872/fathom/src/core.rs#L707-L716

fathom/src/driver/files.rs
```diff
@@ -0,0 +1,65 @@
+//! A reimplementation of `codespan-reporting::files::SimpleFiles` that uses
+//! `FileId` as the file id, instead of `usize`.
+
+use std::ops::Range;
+
+use codespan_reporting::files::{Error, Files, SimpleFile};
+
+use crate::source::FileId;
```
brendanzab, line 8: I think we could move this into the `files` module.

fathom/src/driver/files.rs
```diff
@@ -0,0 +1,65 @@
...
+
+pub struct SimpleFiles<Name, Source> {
```
brendanzab, line 10: Let’s call this `Files`.

fathom/src/source.rs
```diff
@@ -77,7 +80,9 @@ impl StringInterner {
 }
 
 /// File id.
-pub type FileId = usize; // TODO: use wrapper struct
+// 4 billion files should be enough for anyone, and `u16` doesn't save any size in `ByteRange` or `Span`.
+// `NonZeroU32` saves 4 bytes on the size of `Span` compared to `u32`.
+pub type FileId = NonZeroU32; // TODO: use wrapper struct
```
brendanzab, line 85: Nice! I like the use of a `NonZeroU32`. I think it’d be nice to use a wrapper struct here to hide the internal representation.

fathom/src/source.rs
```diff
@@ -172,6 +183,13 @@ pub struct ByteRange {
     end: BytePos,
 }
 
+#[test]
+
```
brendanzab, line 187: Nit: space between the attribute and the definition (not sure if there’s a rustfmt option for this?)

fathom/src/driver.rs
```diff
@@ -11,6 +11,8 @@ use crate::source::{ByteRange, FileId, Span, StringInterner};
 use crate::surface::{self, elaboration};
 use crate::BUG_REPORT_URL;
 
+mod files;
```
brendanzab, line 14: I can see the idea behind putting this under `driver`, but I think it might be better off as a top-level module for now – especially if we decide to move `FileId` into the `files` module (see later review comment).

brendanzab (changes requested):
Really great stuff, thanks!

fathom/src/core.rs
```diff
@@ -713,4 +713,10 @@ mod tests {
         assert!(!std::mem::needs_drop::<Term<'_>>());
         assert!(!std::mem::needs_drop::<Term<'_>>());
     }
+
+    #[test]
+    #[cfg(target_pointer_width = "64")]
+    fn term_size() {
+        assert_eq!(std::mem::size_of::<Term>(), 56);
+    }
```
brendanzab, line 721: Lol I had no idea you could put tests on impls 🤯 - I guess it makes sense? Still, would probably prefer it to live in a test module, but it’s not a huge issue.

brendanzab (approved):

PR #447, opened 2023-01-11
Rename `ByteRange` to `FileRange`

Kmeakin:
Rename `ByteRange` to `FileRange`, to better convey that every AST node carries its `file_id`.
Introduce a separate `ByteRange` with no `file_id`, in case we want to remove `file_id`s from some/all AST nodes later to save space/allocations.


brendanzab:
Closed by #456

PR #449, opened 2023-01-11
Use implicit parameters in primitives and opentype format

brendanzab:
This uses implicit parameters to clean up some functions in the opentype format description, and updates the primitives to make use of implicit parameters.

### DIFF
--- a/formats/opentype.fathom
+++ b/formats/opentype.fathom
@@ -88,7 +88,7 @@ def table_directory (file_start : Pos) = {
             fun (table_id : Repr tag) =>
             fun (table_format : Format) => {
                 // TODO: let formats
-                table_record <- unwrap (Repr table_record) (find_table table_records table_id),
+                table_record <- unwrap (find_table table_records table_id),
                 link <- link_table file_start table_record table_format,
             };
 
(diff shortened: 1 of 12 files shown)


formats/opentype.snap
```diff
@@ -16,8 +16,12 @@ def find_table : fun (num_tables : U16) -> Array16 num_tables {
...
     offset : U32,
     length : U32,
-} =
-fun num_tables table_records table_id => array16_find num_tables (Repr table_record) (fun table_record => table_record.table_id == table_id) table_records;
+} = fun @num_tables table_records table_id => array16_find @num_tables @{
+    table_id : U32,
+    checksum : U32,
+    offset : U32,
+    length : U32,
+} (fun table_record => table_record.table_id == table_id) table_records;
```
brendanzab, line 24: Note that this results bloated terms in the elaborated output. I’d imagine that glued evaluation _might_ be able to keep these terms small, but I could be wrong.

formats/opentype.fathom
```diff
@@ -291,15 +291,15 @@ def table_record = {
...
     fun (table_id : Repr tag) =>
         // TODO: accelerate using binary search
         // TODO: make use of `table_record.search_range`
         // TODO: make use of `table_record.entry_selector`
         // TODO: make use of `table_record.range_shift`
-        array16_find _ (Repr table_record)
-            (fun table_record => table_record.table_id == table_id)
+        array16_find
+            (fun (table_record : Repr table_record) => table_record.table_id == table_id)
             table_records;
```
brendanzab, line 303: Not sure if postponement would help with avoiding the need for a type annotation here? The issue is that we don’t know how to elaborate `table_record.table_id` without knowing the type of `table_record` parameter before hand which comes from the `table_records` argument that is supplied later.

formats/opentype.fathom
```diff
@@ -88,24 +88,24 @@ def table_directory (file_start : Pos) = {
...
-                table_record <- unwrap (Repr table_record) (find_table _ table_records table_id),
+                table_record <- unwrap (find_table table_records table_id),
                 link <- link_table file_start table_record (table_format table_record.length),
             };
 
         let optional_table =
             fun (table_id : Repr tag) =>
             fun (table_format : Format) =>
-                option_fold (Repr table_record) Format {}
+                option_fold ({} : Format)
```
brendanzab, line 106: It would be nice to postpone the elaboration of `{}` and `()` until later. At the moment we eagerly synthesise them to record literals.

brendanzab:
This should be ready to merge – but it highlights some areas where our elaboration of implicit arguments could be improved.

PR #450, opened 2023-01-11
Document implicit arguments in the reference

Kmeakin:


doc/reference.md
```diff
@@ -726,6 +727,25 @@ let id =
...
+### Implicit arguments
+
+Function parameters can be made *implicit* by prefixing them with `@`:
+
+```fathom
+fun (@A : Type) => A
+```
+
+Implicit arguments will be filled in automatically when they can be inferred by the types of other terms:
+```fathom
```
brendanzab, line 739: Nit: a space between the code fences would be nice here!

doc/reference.md
```diff
@@ -726,6 +727,25 @@ let id =
...
+let id = fun (@A : Type) (a : A) => a;
+id false
+```
+
+Implicit arguments can be passed explictly by prefixing the argument with `@`:
+```fathom
```
brendanzab, line 745: Nit: as above

brendanzab (changes requested):
Not sure whether to call these ‘implicit arguments’ or ‘implicit parameters’… Agda [uses the former](https://agda.readthedocs.io/en/latest/language/implicit-arguments.html), but I’ve been more carefully [distinguishing between arguments and parameters](https://stackoverflow.com/questions/156767/whats-the-difference-between-an-argument-and-a-parameter) (following Rust’s lead, for better or worse) so I’m not sure. I think I tend to lean towards ‘implicit parameter’. It’s not a massive issue though and I won’t hold this PR up on that.

Just a minor formatting thing, but otherwise looks good. Thanks!

brendanzab (approved):

doc/reference.md
```diff
@@ -39,6 +39,7 @@ elaboration, and core language is forthcoming.
   - [Function types](#function-types)
   - [Function literals](#function-literals)
   - [Function applications](#function-applications)
+  - [Implicit arguments](#implicit-arguments)
```
brendanzab, line 42: Oh this seems to have been forgotten 😅

brendanzab (changes requested):
The link in the contents just needs to be updated, otherwise we’re good to go.

brendanzab (approved):

PR #451, opened 2023-01-11
Add `absurd`, the eliminator for `Void`

Kmeakin:
Felt we should have this for completeness, though of course it should be impossible to use.


brendanzab:
Thanks!

PR #453, opened 2023-01-12
Implicit function cleanups

brendanzab:
Some cleanups to the implementation of implicit functions that were added in #439. I’ve broken the dependency on `core::Plicity` in the surface syntax, adjusted some names, and cleaned up the insertion of implicit arguments somewhat (inspired by [elaboration-zoo’s implementation](https://github.com/AndrasKovacs/elaboration-zoo/blob/master/04-implicit-args/Elaboration.hs)).

### DIFF
--- a/fathom/src/surface.rs
+++ b/fathom/src/surface.rs
@@ -68,7 +68,7 @@ pub struct ItemDef<'arena, Range> {
     /// The label that identifies this definition
     label: (Range, StringId),
     /// Parameter patterns
-    params: &'arena [FunParam<'arena, Range>],
+    params: &'arena [Param<'arena, Range>],
     /// An optional type annotation for the defined expression
     r#type: Option<&'arena Term<'arena, Range>>,
     /// The defined expression
@@ -212,20 +212,20 @@ pub enum Term<'arena, Range> {
     /// Dependent function types.
     FunType(
         Range,
-        &'arena [FunParam<'arena, Range>],
+        &'arena [Param<'arena, Range>],
         &'arena Term<'arena, Range>,
     ),
     /// Function literals.
     FunLiteral(
         Range,
-        &'arena [FunParam<'arena, Range>],
+        &'arena [Param<'arena, Range>],
         &'arena Term<'arena, Range>,
     ),
     /// Applications.
     App(
         Range,
         &'arena Term<'arena, Range>,
-        &'arena [AppArg<'arena, Range>],
+        &'arena [Arg<'arena, Range>],
     ),
     /// Dependent record types.
     RecordType(Range, &'arena [TypeField<'arena, Range>]),
@@ -333,15 +333,14 @@ impl<'arena> Term<'arena, ByteRange> {
 }
 
 #[derive(Debug, Clone)]
-pub struct FunParam<'arena, Range> {
+pub struct Param<'arena, Range> {
     pub plicity: Plicity,
     pub pattern: Pattern<Range>,
     pub r#type: Option<Term<'arena, Range>>,
 }
 
 #[derive(Debug, Clone)]
-
-pub struct AppArg<'arena, Range> {
+pub struct Arg<'arena, Range> {
     pub plicity: Plicity,
     pub term: Term<'arena, Range>,
 }
--- a/fathom/src/surface/distillation.rs
+++ b/fathom/src/surface/distillation.rs
@@ -59,10 +59,10 @@ impl<'interner, 'arena, 'env> Context<'interner, 'arena, 'env> {
     }
 
     fn push_local(&mut self, name: Option<StringId>) -> StringId {
-        let name =
-            name.unwrap_or_else(|| {
-                self.interner.borrow_mut().get_or_intern_static("_") // TODO: choose a better name?
-            });
+        let name = name.unwrap_or_else(|| {
+            // TODO: choose a better name?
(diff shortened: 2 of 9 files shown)


fathom/src/surface/distillation.rs
```diff
@@ -275,10 +274,9 @@ impl<'interner, 'arena, 'env> Context<'interner, 'arena, 'env> {
                 let body_expr = self.check(body_expr);
                 self.truncate_local(initial_local_len);
 
-                let params = params.into_iter().map(|(plicity, name)| FunParam {
-                    plicity,
-                    pattern: Pattern::Name((), name),
-                    r#type: None,
+                let params = params.into_iter().map(|(plicity, name)| match plicity {
+                    Plicity::Explicit => Param::Explicit(Pattern::Name((), name), None),
```
Kmeakin, line 278: Refactor this into a function?
eg
```rust
impl Param {
    pub fn new(plicity: Plicity, name: Pattern, r#type: Term) -> Self {...}
}
```

fathom/src/surface/distillation.rs
```diff
@@ -471,35 +468,38 @@ impl<'interner, 'arena, 'env> Context<'interner, 'arena, 'env> {
...
                                 self.scope.to_scope(body_type),
                             );
                         }
+                        // Use a parameter if it is referenced in the body
+                        core::Term::FunType(_, plicity, param_name, param_type, next_body_type) => {
+                            let param_type = self.check(param_type);
+                            let param_name = self.freshen_name(*param_name, next_body_type);
+                            let param_name = self.push_local(param_name);
+                            params.push(match plicity {
+                                Plicity::Explicit => {
```
Kmeakin, line 494: As above, refactor into `Param::new()` function

fathom/src/surface/distillation.rs
```diff
@@ -531,10 +531,9 @@ impl<'interner, 'arena, 'env> Context<'interner, 'arena, 'env> {
                 let body_expr = self.synth(body_expr);
                 self.truncate_local(initial_local_len);
 
-                let params = params.into_iter().map(|(plicity, name)| FunParam {
-                    plicity,
-                    pattern: Pattern::Name((), name),
-                    r#type: None,
+                let params = params.into_iter().map(|(plicity, name)| match plicity {
```
Kmeakin, line 534: Ditto

fathom/src/surface/distillation.rs
```diff
@@ -557,9 +556,9 @@ impl<'interner, 'arena, 'env> Context<'interner, 'arena, 'env> {
 
                 while let core::Term::FunApp(_, plicity, next_head_expr, arg_expr) = head_expr {
                     head_expr = next_head_expr;
-                    args.push(AppArg {
-                        plicity: *plicity,
-                        term: self.check(arg_expr),
+                    args.push(match plicity {
```
Kmeakin, line 559: Ditto

Kmeakin (commented):
Why replace turn `surface::Param` into an enum? What if we want to add other annotations to parameters, such as erasure?

fathom/src/surface/elaboration.rs
```diff
@@ -926,6 +926,28 @@ impl<'interner, 'arena> Context<'interner, 'arena> {
...
     }
 
+    /// Elaborate a list of parameters, pushing them onto the context.
+    fn synth_and_push_params(
+        &mut self,
+        params: &[Param<ByteRange>],
+    ) -> Vec<(ByteRange, Plicity, Option<StringId>, core::Term<'arena>)> {
+        self.local_env.reserve(params.len());
+
+        Vec::from_iter(params.iter().map(|param| {
```
Kmeakin, line 936: Why `Vec::from_iter` over the more familiar `.collect()`?

fathom/src/surface.rs
```diff
@@ -205,27 +204,26 @@ pub enum Term<'arena, Range> {
     /// Arrow types.
     Arrow(
         Range,
-        Plicity,
```
Kmeakin, line 208: Why remove implicit arrow types?

brendanzab: Eventually I was hoping that we could make implicit arguments named. I’m also not sure how useful implicit arrows are? Seeing as you can’t infer them based on other arguments or the return type?

fathom/src/surface/elaboration.rs
brendanzab: It formats nicer hah.

fathom/src/surface.rs
Kmeakin: Hmm, true. Since the most common use case for implicit parameters are type parameters, unnamed type parameters wouldn't be very useful.

fathom/src/surface/elaboration.rs
Kmeakin: :shrug: personal preference. I don't mind either way

brendanzab:
> Why replace turn `surface::Param` into an enum? What if we want to add other annotations to parameters, such as erasure?

I guess I still had in my head the `{A : Type} -> B` syntax, where the “implicitness” is part of the parameter syntax – but as we have said before - this is ambiguous with our record syntax. We could make the erasure annotations fields on the variants… but I do see how it would be a similar problem to implicitness – where we might want to use one of those variants in the surface module? I can re-think that commit if it’s easier.

PR #456, opened 2023-01-13
Use `ByteRange` instead of `FileRange` in surface AST

Kmeakin:
Replaces `FileRange` (`ByteRange` + `FileId`) with `ByteRange` in all surface AST types produced by parsing and consumed by elaboration. This shaves 8 bytes off each of `surface::Term` and `surface::Pattern`, and only makes elaboration slightly more complicated, so ought to be a good performance win.

# Size changes:
* `surface::Term<ByteRange>`: 56 bytes to 48 bytes
* `surface::Pattern<ByteRange>`: 16 bytes to 12 bytes

### DIFF
--- a/fathom/src/driver.rs
+++ b/fathom/src/driver.rs
@@ -9,6 +9,7 @@ use codespan_reporting::term::termcolor::{BufferedStandardStream, ColorChoice, W
 use crate::core::binary::{self, BufferError, ReadError};
 use crate::files::{FileId, Files};
 use crate::source::{ByteRange, Span, StringInterner};
+use crate::surface::elaboration::ItemEnv;
 use crate::surface::{self, elaboration};
 use crate::{core, BUG_REPORT_URL};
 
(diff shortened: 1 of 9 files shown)


fathom/src/driver.rs
```diff
@@ -277,7 +277,8 @@ impl<'surface, 'core> Driver<'surface, 'core> {
 
         let initial_buffer = binary::Buffer::from(buffer_data);
         let mut binary_context = binary::Context::new(initial_buffer);
-        let mut elab_context = elaboration::Context::new(&self.interner, &self.core_scope);
+        let mut elab_context =
+            elaboration::Context::new(format_file_id, &self.interner, &self.core_scope);
```
brendanzab, line 281: Hmm, wouldn’t this be the `module_file_id`?

brendanzab (commented):
Oh yeah, I’ve done something like this in the past (having a file id on the elaboration context). It does get a little strange when thinking about how multi-file elaboration fits in. Not sure the best approach there?

Regarding performance, the bigger wins for the binary interpreter is probably going to be related to figuring out how to avoid including spans in values while still allowing us to have decent error messages when the binary interpreter encounters a parse error.

fathom/src/driver.rs
Kmeakin: It was a hack to ensure `item_env` was shared between the two `elab_context`s. `file_id` had to be initialized to *something* when `elab_context` was created, but it would immediately be overwritten by `elab_module`. I have thought of a more elegant way of handling it now though.

Kmeakin:
> It does get a little strange when thinking about how multi-file elaboration fits in. Not sure the best approach there?

Multi-file elaboration should be fine: use a separate `elab_context` for each file/module, and share `item_env` (and any other global state` between them.

PR #458, opened 2023-01-16
Avoid item names

brendanzab:
Noticed that the item names weren’t being checked during fresh name generation.

### DIFF
--- a/tests/succeed/distillation/fresh-names.snap
+++ b/tests/succeed/distillation/fresh-names.snap
@@ -1,0 +1,12 @@
+stdout = '''
+def a : () : Type = ();
+def const1 : fun (@A : Type) (@B : Type) -> A -> B -> A =
+fun @b @c x y => let x1 : b = x;
+let y1 : c = y;
+x;
+def const2 : fun (@A : Type) (@B : Type) -> A -> B -> A = let b : () = ();
+fun @c @d x y => let x1 : c = x;
+let y1 : d = y;
+x;
+'''
+stderr = ''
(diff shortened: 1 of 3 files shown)


Kmeakin:
Could you add some tests to demonstrate cases where fresh local names were clashing with item names?

brendanzab:
> Could you add some tests to demonstrate cases where fresh local names were clashing with item names?

Good idea! Was a bit lazy.

tests/succeed/distillation/fresh-names.snap
```diff
@@ -0,0 +1,12 @@
+stdout = '''
+def a : () : Type = ();
+def const1 : fun (@A : Type) (@B : Type) -> A -> B -> A =
+fun @b @c x y => let x1 : b = x;
```
brendanzab, line 4: I wonder if we could actually pull these binder names from the type somewhere?

Kmeakin: What do you mean?

tests/succeed/distillation/fresh-names.snap
```diff
@@ -0,0 +1,12 @@
+stdout = '''
+def a : () : Type = ();
```
Kmeakin, line 2: Looks like the type annotation on this item is being distilled in check mode instead of synth mode?

brendanzab: Oooh, good catch

tests/succeed/distillation/fresh-names.snap
brendanzab: Like, try to use `A` and `B`. Just a thought though!

PR #459, opened 2023-01-16
Distill metavariables to named terms

Kmeakin:
Distills `core::Term::MetaVar` to `surface::Term::Name`, where the name is of the form `?{level}`. Since metavariables are global per module, we don't need to generate fresh names.


fathom/src/surface/distillation.rs
```diff
@@ -407,7 +407,10 @@ impl<'interner, 'arena, 'env> Context<'interner, 'arena, 'env> {
             },
             core::Term::MetaVar(_span, var) => match self.get_hole_name(*var) {
                 Some(name) => Term::Hole((), name),
-                None => Term::Placeholder(()),
+                None => Term::Name(
+                    (),
+                    self.interner.borrow_mut().get_or_intern(format!("?{var}")),
+                ),
```
brendanzab, line 413: What about updating `get_hole_name` to return `format!("{var}")` if the metavariable is unnamed?

PR #460, opened 2023-01-17
Suggest a similar name when reporting UnboundName or UnknownField

Kmeakin:
Include the most similar name in scope when reporting `UnboundName`/most similar label when reporting `UnknownField`, according to their Levenshtein distance.

Will probably need to fine tune some heuristics in the future:
* Should we only report a suggestion if the Levenshtein distance is below a certain limit?
* Should we report several potential suggestions, not just one?


fathom/src/surface/elaboration.rs
```diff
@@ -283,6 +283,22 @@ pub struct Context<'interner, 'arena> {
...
+) -> Option<StringId> {
+    let name = interner.resolve(name).unwrap();
+
+    candidates
+        .map(|symbol| {
+            let other_name = interner.resolve(symbol).unwrap();
+            (levenshtein::levenshtein(name, other_name), symbol)
+        })
+        .min_by_key(|(weight, _)| *weight)
+        .map(|(_, symbol)| symbol)
```
brendanzab, line 299: Does this work?

```suggestion
    candidates.min_by_key(|candidate| {
        let candidate = interner.resolve(candidate).unwrap();
        levenshtein::levenshtein(name, candidate)
    })
```

brendanzab (changes requested):
Sweet! Thanks for this!

PR #462, opened 2023-01-17
Avoid unnecessary quotation in elaboration

brendanzab:
This reduces some bloat in the elaboration output

### DIFF
--- a/tests/succeed/primitive-ops.snap
+++ b/tests/succeed/primitive-ops.snap
@@ -1,12 +1,12 @@
 stdout = '''
-let test : Array8 3 () -> Array8 3 () = fun x => x;
-let test : Array16 3 () -> Array16 3 () = fun x => x;
-let test : Array32 3 () -> Array32 3 () = fun x => x;
-let test : Array64 3 () -> Array64 3 () = fun x => x;
-let test : Array8 2 () -> Array8 2 () = fun x => x;
-let test : Array16 2 () -> Array16 2 () = fun x => x;
-let test : Array32 2 () -> Array32 2 () = fun x => x;
-let test : Array64 2 () -> Array64 2 () = fun x => x;
+let test : Array8 ((1 : U8) + (2 : U8)) () -> Array8 3 () = fun x => x;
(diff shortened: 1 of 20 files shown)


tests/succeed/primitive-ops.snap
```diff
@@ -1,12 +1,12 @@
 stdout = '''
-let test : Array8 3 () -> Array8 3 () = fun x => x;
-let test : Array16 3 () -> Array16 3 () = fun x => x;
-let test : Array32 3 () -> Array32 3 () = fun x => x;
-let test : Array64 3 () -> Array64 3 () = fun x => x;
-let test : Array8 2 () -> Array8 2 () = fun x => x;
-let test : Array16 2 () -> Array16 2 () = fun x => x;
-let test : Array32 2 () -> Array32 2 () = fun x => x;
-let test : Array64 2 () -> Array64 2 () = fun x => x;
+let test : Array8 ((1 : U8) + (2 : U8)) () -> Array8 3 () = fun x => x;
```
Kmeakin, line 2: I think we want simple expressions like this to be reduced in the final output. Maybe allow marking items with `#[opaque]` to prevent them reducing during quoting or something?

brendanzab: I dunno - isn’t this reflecting the elaboration input more accurately though? i.e.
https://github.com/yeslogic/fathom/blob/0b233a745d5176c0bf2668f823fc7f732b23f967/tests/succeed/primitive-ops.fathom#L1

Kmeakin: If I write something like
```fathom
let x : id Bool = false;
x
```
I would expect the type of `x` to be displayed as `Bool`, not `id Bool`

brendanzab: I think in type errors this would still happen though, as we generally evaluate types before displaying them to the user.

Kmeakin: Hmm, I need to think about it more. I do like how it makes the output in the other cases more readable!

brendanzab: Another thing I’d like to look into at some stage is glued evaluation which is related to this stuff IIUC?

PR #464, opened 2023-01-18
Use `Iterator::try_fold` instead of `fold` when renaming terms

Kmeakin:
Encountered this while writing my own unifier (https://github.com/Kmeakin/pion/blob/master/crates/pion-core/src/elab/unify.rs#L415). `try_fold` is short circuiting (will return early on the first `Err`), while `fold` is not.


Kmeakin:
Submitted an issue to detect this pattern with clippy: https://github.com/rust-lang/rust-clippy/issues/10208

brendanzab:
Ooh, very handy!

PR #465, opened 2023-01-18
Introduce a smart constructor around `String` for input sources

Kmeakin:
Adds `ProgramSource`, a wrapper around `String` which guarantees the length is `<= u32::MAX`. Allows us to remove the `assert` from `tokens`.


brendanzab (approved):
Looks good!

PR #466, opened 2023-01-19
Update `scoped-arena` from `0.3` to `0.4.1`

Kmeakin:


brendanzab:
Were the changes to `alloc.rs` related to the upgrade? If so, then it might help to describe why the changes were made in the PR description. If the weren’t related to the upgrade then it might be better to split them out into a separate commit with an appropriate description. I do like the changes though!

brendanzab:
Thanks for the update at any rate!

Kmeakin:
> Were the changes to `alloc.rs` related to the upgrade? If so, then it might help to describe why the changes were made in the PR description. If the weren’t related to the upgrade then it might be better to split them out into a separate commit with an appropriate description. I do like the changes though!

The only change strictly related to the update was using `scope.to_many_with`, which was added in version `0.4.0`

PR #468, opened 2023-01-19
Slicevec cleanup

Kmeakin:
- Rename `next_index` to `len`
- Rename `max_len` to `capacity`
- Add accessor fns `len()`, `capacity()` and `is_full()`
- Panic with an informative message when trying to push onto a full `SliceVec`


brendanzab (approved):
Nice cleanup!

PR #473, opened 2023-01-21
Replace all spellings of `-ise` with `-ize`

Kmeakin:
We alternated between the [American `-ize` and British/Commonwealth `-ise`](https://en.wikipedia.org/wiki/American_and_British_English_spelling_differences#-ise,_-ize_(-isation,_-ization)) spelling inconsistently. I propose we standardize on the American spelling, in line with the Rust standard library documentation. [RFC 1574](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md) states:

> All documentation for the standard library is standardized on American English, with regards to spelling, grammar, and punctuation conventions. Language changes over time, so this doesn’t mean that there is always a correct answer to every grammar question, but there is often some kind of formal consensus.

### DIFF
--- a/fathom/src/core/semantics.rs
+++ b/fathom/src/core/semantics.rs
@@ -1,5 +1,5 @@
-//! The semantics of the core language, implemented using [normalisation by
-//! evaluation](https://en.wikipedia.org/wiki/Normalisation_by_evaluation).
+//! The semantics of the core language, implemented using [normalization by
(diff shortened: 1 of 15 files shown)


brendanzab:
Ahh I guess that’s fair. Struggle a bit with my commonwealth english, hah 😳

PR #474, opened 2023-01-21
Make parenthesized terms explicit in the surface AST

Kmeakin:
This would allow us to pretty print the user's input exactly in the future (eg if we want a `fathom fmt` command). Would also allow us to move the precedence logic from `surface::pretty` to `core::distillation`


brendanzab:
Good idea!

PR #476, opened 2023-01-23
Lazy evaluation

Kmeakin:


fathom/src/core/semantics.rs
```diff
@@ -97,7 +146,7 @@ pub enum Head {
 #[derive(Debug, Clone)]
 pub enum Elim<'arena> {
     /// Function applications.
-    FunApp(Plicity, ArcValue<'arena>),
+    FunApp(Plicity, LazyValue<'arena>),
```
brendanzab, line 149: Could also use `LazyValue` in the parameter types of `Value::FunType` and `Value::FormatCond`.

fathom/src/core/semantics.rs
```diff
@@ -52,11 +59,58 @@ pub enum Value<'arena> {
...
+/// Initialization operation for lazy values.
+///
+/// We need to use a [defunctionalized] representation because Rust does not
+/// allow closures of type `dyn (Clone + FnOnce() -> Arc<Value>)`.
+///
+/// [defunctionalized]: https://en.wikipedia.org/wiki/Defunctionalization
+#[derive(Clone, Debug)]
+enum LazyInit<'arena> {
+    EvalTerm(SharedEnv<ArcValue<'arena>>, &'arena Term<'arena>),
+    ApplyElim(Arc<LazyValue<'arena>>, Arc<Elim<'arena>>),
```
brendanzab, line 71: Why do we need `Arc`s for `LazyValue` and `Elim`? Is this to keep the size down?

brendanzab:
Very nice, I’m happy with this. Is there anything else that needs to be done to the before merging?

fathom/src/core/semantics.rs
Kmeakin: Both are necessary to prevent recursion/infinite size error

Kmeakin:
Should types of items/metavars/local vars be lazily evaluated too?

PR #479, opened 2023-01-27
Use explicit type names in place of `Self` where possible

Kmeakin:
Use explicit type names in place of `Self` where possible. Dual to https://github.com/yeslogic/fathom/pull/480


brendanzab:
Let’s go with this - asking others it seems like there’s not a lot of consistency - some people using `Self` where if makes things clearer, and others in constructors but sometimes `Self` return types.

PR #480, opened 2023-01-27
Use `Self` in place of explicit type names where possible

Kmeakin:
Use `Self` in place of explicit type names where possible. 
Also adds `#![warn(clippy::use_self)]` to enforce the new style. 

Dual to https://github.com/yeslogic/fathom/pull/479


brendanzab:
Closed by #479

PR #482, opened 2023-01-27
Add `CollectIntoScope` trait

Kmeakin:
Added the CollectIntoScope::collect_into_scope method, analogous to `Iterator::collect`. Allows postfix iterator chaining, which may make code more readable.

If it gets merged upstream (see https://github.com/zakarumych/scoped-arena/pull/2), can remove the implementation later

### DIFF
--- a/fathom/src/alloc.rs
+++ b/fathom/src/alloc.rs
@@ -115,3 +115,17 @@ pub unsafe fn slice_assume_init_ref<'a, T>(slice: &'a [MaybeUninit<T>]) -> &'a [
     // valid for reads.
     &*(slice as *const [MaybeUninit<T>] as *const [T])
 }
+
+/// An extension trait that provides a postfix version of
+/// `Scope::to_scope_from_iter`. This may lead to more readable code in some
+/// instances.
+pub trait CollectIntoScope<T> {
+    #[allow(clippy::mut_from_ref)]
+    fn collect_into_scope<'a>(self, scope: &'a scoped_arena::Scope<'a>) -> &'a mut [T];
+}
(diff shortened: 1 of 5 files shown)


brendanzab:
Neat, seems like it was accepted!

Yeah have wondered about making allocating terms nicer. I wonder, for example, if we could:

- consider breaking the allocator out of the context, instead passing it as an argument to things that need it (this might help with lifetime juggling or might not, I dunno)
- investigate a specific struct or trait for building nodes: Chalk has an [`Intern` trait](https://rust-lang.github.io/chalk/book/types/role_of_interner.html) for allocating terms as well which seems interesting

Not sure though - but yeah I get a bit sick of `self.scope.to_scope` and not being able to nest calls to certain functions due to having a mutable reference to the context.

Kmeakin:
> Neat, seems like it was accepted!

Should we update our dependency to the latest commit, or wait for another `crates.io` release?

brendanzab:
I guess I think it‘s a bit of a bummer that things flow onto multiple lines… not sure if that is an improvement to readability or not. It’s a pretty common struggle with `Iterator::collect` too, alas.

Kmeakin:
> I guess I think it‘s a bit of a bummer that things flow onto multiple lines… not sure if that is an improvement to readability or not. It’s a pretty common struggle with `Iterator::collect` too, alas.

Might be possible to alter the layout with some rustfmt configuration option?

brendanzab:
Not 100% sure this is an improvement due to the line reflows… so just gonna close this for now. Thanks for the experiment though!

PR #483, opened 2023-01-28
Optimise `surface::pretty`

Kmeakin:
Fixes very high heap utilization when rendering `surface` AST to `Doc` for pretty printing.

Problem was in `surface::pretty::sequence`, in particular the use of `FlatAlt`. `FlatAlt` requires allocating two AST nodes, one for the single line situation and one for the multi line situation. When used recursively, this will lead to exponential space and time complexity! Fortunately, we can push `FlatAlt` down into the leaves: it is only necessary when deciding whether to emit a separator (eg `,`) or `nil` at the end of the item in the list.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,7 @@
+[profile.release-debug]
+inherits = "release"
+debug = true
+
 [workspace]
 members = [
     './fathom',
--- a/fathom/src/surface/pretty.rs
+++ b/fathom/src/surface/pretty.rs
@@ -179,33 +179,44 @@ impl<'interner, 'arena> Context<'interner, 'arena> {
                     else_expr = next_else;
                 }
 
-                if branches.is_empty() {
-                    self.pretty_if(cond_expr, then_expr, else_expr)
-                } else {
-                    self.pretty_if_else_chain(cond_expr, then_expr, branches, else_expr)
-                }
+                self.text("if ")
+                    .append(self.term(cond_expr))
+                    .append(
+                        self.line()
+                            .append(self.text("then "))
+                            .append(self.term(then_expr))
+                            .nest(INDENT),
+                    )
+                    .append(self.concat(branches.iter().map(|(cond_expr, then_expr)| {
+                        self.line()
+                            .append("else if ")
+                            .append(self.term(cond_expr))
+                            .append(" then ")
+                            .append(self.term(then_expr))
+                            .nest(INDENT)
+                    })))
+                    .append(
+                        self.line()
+                            .append(self.text("else "))
+                            .append(self.term(else_expr))
+                            .nest(INDENT),
+                    )
+                    .group()
             }
-            Term::Match(_, scrutinee, equations) => self.sequence(
-                true,
-                self.concat([
-                    self.text("match"),
-                    self.space(),
-                    self.term(scrutinee),
-                    self.space(),
-                    self.text("{"),
-                ]),
-                equations.iter().map(|(pattern, body_expr)| {
-                    self.concat([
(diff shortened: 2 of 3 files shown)


Kmeakin:
# Benchmark results

## Before
Memory usage, as measured by `heaptrack ./target/release/fathom data --module formats/opentype.fathom /usr/share/fonts/liberation/LiberationMono-Bold.ttf`: 142.9MB

Runtime, as measured by `hyperfine -w 1 -r 10 "./target/release/fathom data --module formats/opentype.fathom /usr/share/fonts/liberation/LiberationMono-Bold.ttf"`:
```
  Time (mean ± σ):     341.7 ms ±   3.3 ms    [User: 315.4 ms, System: 25.9 ms]
  Range (min … max):   337.2 ms … 347.1 ms    10 runs
```

## After
Memory usage, as measured by `heaptrack ./target/release/fathom data --module formats/opentype.fathom /usr/share/fonts/liberation/LiberationMono-Bold.ttf`: 14.2MB

Runtime, as measured by `hyperfine -w 1 -r 10 "./target/release/fathom data --module formats/opentype.fathom /usr/share/fonts/liberation/LiberationMono-Bold.ttf"`:
```
  Time (mean ± σ):      38.0 ms ±   1.0 ms    [User: 33.9 ms, System: 4.2 ms]
  Range (min … max):    37.1 ms …  39.7 ms    10 runs
```

Cargo.toml
```diff
@@ -1,3 +1,6 @@
+[profile.release]
+debug = true
+
```
brendanzab, line 3: Perhaps we could use a [custom profile](https://doc.rust-lang.org/cargo/reference/profiles.html#custom-profiles):

```suggestion
# Release profile with debug symbols, useful for profiling
[profile.release-debug]
inherits = "release"
debug = true
```

That way debug symbols are only generated if desired.

brendanzab (changes requested):
Awesome: good catch, thanks!

fathom/src/surface/pretty.rs
```diff
@@ -576,19 +445,66 @@ impl<'interner, 'arena, A: 'arena> DocAllocator<'arena, A> for Context<'interner
                     RefDoc(&Doc::Nil),
                 )))
             }
+            _ => self.scope.to_scope(doc),
```
brendanzab, line 448: Hmm, this seems to be breaking clippy?

fathom/src/surface.rs
```diff
@@ -145,6 +145,21 @@ impl<Range> BinOp<Range> {
...
+            BinOp::Mul(_) => " * ",
+            BinOp::Div(_) => " / ",
+            BinOp::Eq(_) => " == ",
+            BinOp::Neq(_) => " != ",
+            BinOp::Lt(_) => " < ",
+            BinOp::Lte(_) => " <= ",
+            BinOp::Gt(_) => " > ",
+            BinOp::Gte(_) => " >= ",
+        }
+    }
```
brendanzab, line 161: Can we move this into `pretty`?

fathom/src/surface/pretty.rs
Kmeakin: Was keeping it there temporarily to remind me to investigate it once I was done with everything else. It's been fixed now

brendanzab:
The initial win was good, but I’m not sure about the later micro-optimsations. I think I’d prefer to hold off on these to another PR. I’m somewhat uncomfortable with the interner optimisation: while it looks tempting it I think I’d prefer it if we could properly encapsulate the unsafe code around the string interner.

PR #492, opened 2023-02-04
Make string interner global

### LINKED ISSUE #491: Global string interner
I think we should make the `StringInterner` global. Usually global variables are seen as a code smell, but I think it can be justified in this case:

* As far as I know, there is no use case for having multiple `StringInterner`s in the same process.
* If that is the case, passing around a single `StringInterner` adds extra complexity and inconvenience for no benefit.
* Since `StringId`s are an integer index, not a pointer, resolving can fail, but there isn't any good recovery strategy (and logically shouldn't be possible), so will always be `unwrap`ed anyway
* Once a string has been interned, it should be live and dereferencable for the rest of the process - ie stored in a global interner

Kmeakin:
Fixes #491


fathom/src/symbol.rs
```diff
@@ -0,0 +1,174 @@
...
+        Self(INTERNER.write().unwrap().strings.get_or_intern_static(sym))
+    }
+
+    pub fn resolve(self) -> &'static str {
+        let interner = INTERNER.write().unwrap();
+        let symbol = interner.strings.resolve(&self.0);
+
+        // SAFETY: `INTERNER` is static, so references into it are `'static`
+        unsafe { std::mem::transmute::<&str, &str>(symbol) }
+    }
```
brendanzab, line 126: Hmm, how sure are we that this is safe and not UB? Can we run it by MIRI? I would at least prefer we used `std::mem::transmute::<&str, &'static str>(symbol)` to be explicit.

brendanzab: I guess I’ve always associated `'static` with stuff that is in the static code of the program, where as the strings in `ThreadedRodeo` are being allocated at runtime. I could be off on this understanding, however. It does give me pause though.

Kmeakin: My understanding is `'static` is simply the lifetime that outlives all others. Usually comes from string literals, but can also come from eg leaking an allocation.

Kmeakin: If you are uncomfortable with the `'static` lifetime, we could change the signature to 
`fn resolve<'a>(&'a self) -> &'a str` (with explicit lifetimes added for clarity).

[Taken from rustc](https://github.com/rust-lang/rust/blob/e4dd9edb76a34ecbca539967f9662b8c0cc9c7fb/compiler/rustc_span/src/symbol.rs#L1845):

```rs
impl Symbol {
    /// Access the underlying string. This is a slowish operation because it
    /// requires locking the symbol interner.
    ///
    /// Note that the lifetime of the return value is a lie. It's not the same
    /// as `&self`, but actually tied to the lifetime of the underlying
    /// interner. Interners are long-lived, and there are very few of them, and
    /// this function is typically used for short-lived things, so in practice
    /// it works out ok.
    pub fn as_str(&self) -> &str {
        with_session_globals(|session_globals| unsafe {
            std::mem::transmute::<&str, &str>(session_globals.symbol_interner.get(*self))
        })
    }
}
```

brendanzab: Ahh, fair. If it’s ok for rustc then it’s ok for me! Let’s change the type signature.

Kmeakin:
Unfortunately tying the lifetime of the string to the lifetime of the symbol means we cant pass the strings by reference to `pretty`, and have to do some unnecessary clones. I'm pretty confident transmuting to `'static` is safe, since `INTERNER` is never dropped.

brendanzab:
Seems like we were doing that anyway at any rate. Lets go with this for now.

PR #494, opened 2023-02-04
Cleanup distillation

Kmeakin:
Miscellaneous cleanups/refactors of `surface::distillation`:

* Replace the mutually recursive `check` and `synth` functions with a single `term` function. Now all logic is contained in one single function, and is IMO easier to understand.
* Boolean patterns were erroneously distilled to name patterns instead of boolean patterns in the surface syntax
* Avoid allocating an intermediate `Vec` when distilling a function application to a binop


brendanzab:
Nice cleanups!

Issue #488, opened 2023-02-02
Implementation annoyances


brendanzab:
Here’s a list of things that are annoying/frustrating in the implementation of Fathom:

- `self.scope.to_scope` and `self.scope.to_scope_from_iter` are very noisy and make elaboration, quotation, etc. difficult to read.
- `self.eval_env().eval(...)`, `self.quote_env().quote(...)` etc. is rather noisy
- can’t allocate AST nodes when contexts are borrowed mutably, meaning we need to split things up over multiple lines
- can’t report errors while and pretty printing terms/values in the same expression sure to multiple mutable borrows of the context
- pushing and popping contexts and environments adds lots of noise and is easy to mess up
- ranges vs. file ranges vs. spans - it’s annoying to have to convert between these
- it’s hard to match on primitives and their arguments
- initialising the types of primitives is verbose and error prone
- initialisation of primitive evaluation functions is clunky

The combined effect of these means that it is harder to implement new features and to experiment with new ideas.

Some potential ideas for improvements (which may or may not help):

- delegate allocator and evaluation/quotation functions in the contexts that need them
- experiment with using immutable environments vs. pushing and popping environments (not sure about the performance impact of this)
- when pushing to an environment, take a closure that then handles resetting the environment back to its original state
- pass the allocator `Scope` as a separate argument
- use allocators that use IDs vs. pointers
- move spans into a side-table that can be referred to by ID
- go back to using `Box`es and and `Vec`s in surface and core syntaxes
- store spans in the surface syntax, and refer back to original surface nodes by pointer/ID in terms/values
- avoid the need to mutably access the list of diagnostics when reporting errors – perhaps by using shared memory
- investigate a “term builder” for defining the types of primitives like in https://gist.github.com/TOTBWF/9b2c071d2edb1c6596b785656c866fd6

We might need to experiment with some of these outside of Fathom before committing to anything. Looking at how other Rust projects handle some of these issues might also be interesting.

Kmeakin:
I would recommend against using integer IDs instead of pointers:
* Have to pass the arena to any code that wants to dereference an ID
* `fmt::Debug` output is less readable because subexpressions are just printed as an opaque ID
* Instead of a single `scoped_arena::Scope`, would need a separate arena for each type being allocated
* Introduces runtime bounds checks when de-referencing
* Possible to index an arena with an ID that was allocated from a different area (eg if each module has its own arena, would be possible to index module 1 with a `TermId` from module 2)

[...]

PR #496, opened 2023-02-09
Make it easier to distill/pretty print values

brendanzab:
In an attempt to address #488, this reduces some of the gymnasics we’ve needed to do whenever we want to pretty print something. We achieve this by cloning the name environment as opposed to taking it by mutable reference when constructing a distillation environment.

This means we can no longer reuse the existing allocation when temporarily pushing local names, but I think the savings we got from this was probably minor, and the increase in clarity is worth it.


wezm (commented):
Looks good 👍

PR #502, opened 2023-02-13
Remove `Array8`, `Array16` and `Array32` types

Kmeakin:
* Removes the `Array8`, `Array16` and `Array32` types and their corresponding functions, leaving only `Array64`.
* Added `<from>_extend_<to>` and `<from>_truncate_<to>` for all integer types 
* ~~Implicitly insert extension functions during coercion. This allows using shorter integers (eg parsed from input) to operate on `Array64` without hassle~~


Kmeakin:
Removed implicit extension because it could silently change the result of expressions (eg `(u16_extend_u64 x * u16_extend_u64 y) != u16_extend_u64 (x * y)`)

